Issue #8, opened 2019-05-27
mst-gql should support a `mst-gql.json` file for configuration


mweststrate:
Instead, or additionally to, using the cli options, which can be pretty long. 

Also, some future options might need richer data than can be easily provided in a CLI.

elie222:
Did you consider using graphql-codegen? It’s a library with type and code generators for lots of languages. Ts, flow, Apollo, urql... 

Cli/json options come out the box with it.

elie222:
https://graphql-code-generator.com/docs/getting-started/

mweststrate:
@elie222 yes, see #13

PR #14, opened 2019-05-27
(feat) - make a basic config

JoviDeCroock:
Hey,

Thought I'd make my attempt at chiming in, marked this as draft to first get some initial thoughts about:
1. the naming
2. what is configurable
3. what gets presedence args or config
4. do we pass a config file to cli like --config x.js or is an opinionated name okay?
5. JSON-support aswell?

Awesome project! Keep up the good work

Implements: https://github.com/mobxjs/mst-gql/issues/8


mweststrate:
1. I'm so bad at that. `.mstgqlrc` would work, but I don't like those hidden files. `mst-gql.json` (also `.js`?) would work I think?
2. For now all the same options as supported by the cli
3. args should override config imho
4. I think it should pick it up automatically from the root dir, but should be possible to provide it with a flag as well, in case there are 2 configs (for example in the twitter example one for the server and a different one for the client)
5. Yeah I would assume json only, or a .js file that exports a similarly shaped object as default when required.

JoviDeCroock:
The `||` approach implemented now isn't maintainable. I'd like to make some sort of "finalConfig" object which is a merge of both of these. First wanted to get your thoughts on this.

I'll implement JSON and will make args have presedence over the config.

mweststrate:
Awesome!

FredyC:
I would suggest using https://github.com/davidtheclark/cosmiconfig instead of reinventing the well. Configs are already hard, this package brings some standardization at least.

JoviDeCroock:
> I would suggest using https://github.com/davidtheclark/cosmiconfig instead of reinventing the wheel. Configs are already hard, this package brings some standardization at least.

I'll look into it, did not know about that. Should've done a bit more research before diving in

mweststrate:
This is still pending correct, or are you already happy about it in it's current form @JoviDeCroock ?

JoviDeCroock:
I haven't found time to look at cosmiconfig yet @mweststrate this will make life easier on you I think, I'll probably find some time this weekend

JoviDeCroock:
Except if you already think it's good enough, then I'll make a seperate PR. I don't really mind, whatever you prefer.

mweststrate:
Separate PR sounds fine by me, and keeps the change more incrementalfor
you, I assume it wouldn't really change behavior.

Op di 4 jun. 2019 20:56 schreef Jovi De Croock <notifications@github.com>:

generator/config.js
```diff
@@ -0,0 +1,60 @@
...
+  modelsOnly: false,
+  outDir: "src/models",
+  roots: [],
+  noReact: false,
+}
+
+exports.getConfig = function getConfig(configPath) {
+  let configLocation
+  if (configPath) {
+    configLocation = resolve(process.cwd(), configPath)
```
mweststrate: Looking through this another, I noticed that relying on `process.cwd` might be tricky, because, if the script is initiated from one of the project's sub directories, it might not find it (afaik an package.json script doesn't automatically have the project root as cwd). But I think we can directly use this package to find the correct file: https://github.com/sindresorhus/find-up

FredyC: Wouldn't be better to go with https://github.com/davidtheclark/cosmiconfig right away which takes care of such edge cases? Feels weird to be reinventing a wheel once again.

mweststrate:
I think in basics this should work fine, so I think it is up to Jovi how
far he wants to take this (or any body else), and in one or 2 steps, but
yes, it seems to exactly address this :)

JoviDeCroock:
@mweststrate I tried the cosmiconfig, I still have to double check things before merging since I'm pretty low on time atm

JoviDeCroock:
Seems to work well, @FredyC I don't know if you have experience with it and maybe see some bad things I'm doing? Would gladly hear so

FredyC:
@JoviDeCroock I only used it as a consumer of various modules, but it seems pretty straightforward, so I wouldn't worry about it much. It can be improved over time if necessary.

JoviDeCroock:
Allright, thanks @FredyC for the recommendation

elie222:
So I haven’t been following this PR too much. Is it now stable for a merge?

elie222:
Ah we have conflicts

elie222:
Merge conflicts :(

JoviDeCroock:
It misses one config option that got added later

Issue #13, opened 2019-05-27
Use a decent code generator or templates


mweststrate:
The current generator `generate.js`, is a pretty straight forward string concatenation tool. it's fast and straight forward, but probably not very maintainable as it grows more complex in the future. Something more solid could be a good idea, I think @birkir had some thoughts about that in the past

mweststrate:
For example: https://github.com/dotansimha/graphql-code-generator

sw-yx:
@jamonholmgren's https://github.com/infinitered/gluegun goes templating well.

jamonholmgren:
Thanks for the mention, @sw-yx!

We're using the `ejs` library to power our template system:

https://github.com/infinitered/gluegun/blob/b35a848c1cf5324f7767d15104416d8a78b2035b/src/toolbox/template-tools.ts#L16

We also have a really cool patching library:

https://github.com/infinitered/gluegun/blob/b35a848c1cf5324f7767d15104416d8a78b2035b/src/toolbox/patching-tools.ts

Both are built-in to Gluegun, which may be overpowered for your needs here. But if you want to make `mst-gql` a full featured CLI, let me know and I can walk you through what you get with Gluegun! Otherwise, feel free to steal/borrow any concepts from the links I posted here.

FredyC:
Should we actually turn this into a monorepo to split stuff into plugins? I am no big fan of Lerna, it does the job, but it also gets in the way fairly often.

mweststrate:
Let's not over engineer yet; it does make contributing harder

Op wo 29 mei 2019 11:10 schreef Daniel K. <notifications@github.com>:

FredyC:
Ok sure, I suppose that Gluegun can use "plugins" from the single package, so that shouldn't matter, but we should also support custom plugins from NPM.

---

Regarding the code generation, I actually do have a custom generator for `graphql-code-generator` which generates fully typed hooks for each graphql operation. I am not sure yet how it would like here, but a similar alternative would be really nice.

```ts
export function useQResolveLocation(
  variables?: QResolveLocationVariables,
  baseOptions?: Hooks.QueryHookOptions<QResolveLocationVariables>,
) {
  return Hooks.useQuery<QResolveLocationQuery, QResolveLocationVariables>(
    QResolveLocationDocument,
    variables,
    baseOptions,
  )
}
```

mweststrate:
Please open separate issues for separate ideas 😉

[...]

PR #17, opened 2019-05-29
feat: added option to not generate react scaffold

### LINKED ISSUE #15: React should be opt-in or opt-out
I think the generation of the code related to React should be an opt-in feature. Store management should be agnostic. 

Or at least, it should be possible to give an **option to not output React code**.

ematipico:
Added an option `--noReact` that prevents from generating react files.

Closes #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,7 @@ The `mst-gql` command currently accepts the following arguments:
 - `--excludes 'type1,type2,typeN'` The types that should be omitted during generation, as we are not interested in for this app.
 - `--roots 'type1,type2,typeN'` The types that should be used as (root types)[#root-types]
 - `--modelsOnly` Generates only models, but no queries or graphQL capabilities. This is great for backend usage, or if you want to create your own root store
+- `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
 - `source` The last argument is the location at which to find the graphQL definitions. This can be
   - a graphql endpoint, like `http://host/graphql`
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -22,7 +22,8 @@ function generate(
   rootTypes = [],
   excludes = [],
   generationDate = "a long long time ago...",
-  modelsOnly = false
+  modelsOnly = false,
+  noReact = false
 ) {
   excludes.push(...buildInExcludes)
 
@@ -40,7 +41,7 @@ function generate(
 
   generateTypes()
   generateRootStore()
-  if (!modelsOnly) {
+  if (!modelsOnly || !noReact) {
     generateReactUtils()
   }
   generateBarrelFile(files)
--- a/generator/mst-gql-scaffold.js
+++ b/generator/mst-gql-scaffold.js
@@ -13,7 +13,8 @@ const definition = {
   "--roots": String,
   "--excludes": String,
   "--modelsOnly": Boolean,
-  "--force": Boolean
+  "--force": Boolean,
+  "--noReact": Boolean
 }
 
 function main() {
@@ -40,6 +41,7 @@ function main() {
     : []
   const modelsOnly = !!args["--modelsOnly"]
   const forceAll = !!args["--force"]
+  const noReact = !!args["--noReact"]
 
   console.log(
     path.basename(__filename) +
@@ -97,7 +99,8 @@ function main() {
     roots,
     excludes,
     new Date().toUTCString(),
-    modelsOnly
+    modelsOnly,
+    noReact
   )
   writeFiles(outDir, files, format, forceAll, true)
 }


dpnolte:
I was wondering if the peer dependency on mobx-react can then also be removed as part of this PR. At first glance, it seems that this peer dep is not necessary anyways? And perhaps react-dom can also be removed?

ematipico:
Yes you are right. Although if the user decides to scaffold the react snippets, somehow we have to notify the user that they have to add these libraries in their dependencies. How should we tackle this?

FredyC:
That's why I was suggesting a plugin architecture in #15. If the actual React bindings would be a plugin, it can have `mobx-react` as a dependency without the need to be in the core package.

ematipico:
Which will be part of #13 and using a better CLI tool such as Gluegun

FredyC:
Well, I don't really see a point in adding some option now when it will become obsolete once a proper tool is used :) It seems that Gluegun indeed supports plugins, so that would solve it.

dpnolte:
@FredyC How do you think that plugin architecture should look like? 
It seems that the plugin would apply mostly to the scaffolding output and the related required dependencies such as react/mobx-react. Additionally, the usQuery hook (react.tsx) needs to be moved to this plugin. Also perhaps nice to have a mobx-react-light plugin.

I guess it would be easiest to add a plugin section to a config file here (see also #8 ).
Using cosmiconfig, it can parse a plugin array that contains a string array of plugin module names.
I was thinking each plugin can than have a function as default export (similar to babel plugins), but with at least the following arguments:
- module name (e.g, TodoModelBase)
- RootStore import path (e.g., './RootStore')
- toExport array 
- generateFile function to add files as output
This should be sufficient to generate the react utils in the plugin.

The scaffolder (generate.js) could then invoke all configured plugins.
I need to have a solution like the above. So if this makes sense, I can give it a go.

FredyC:
@dpnolte That definitely sounds like a good way to go.

Regarding the mobx-react dependency, why is it actually needed? I don't see it used in any generated code. Sure, you would most likely need it in the app to attach an observer to a component, but that's application concern. That means people can use mobx-react-lite if they want just fine.

One approach that could be interesting is wrapping the `useQuery` body into a `useObserver` and that way it wouldn't be required to use observer for the component. It could be an option of the plugin perhaps because in that case you do you need either `mobx-react-lite` or `mobx-react@6`.

mweststrate:
I think for now the peer deps could be removed and documented. I don't
think plugins solve the problem, as it is related to package published to
npm, not the scaffolded code. (Publishing separate package will be a lot of
work, as one would need to align the release, versioning etc. of the
runtime package with the scaffold package)

Op wo 29 mei 2019 11:53 schreef Daniel K. <notifications@github.com>:

dpnolte:
I see, thanks for explaining. Is there also a way to remove the react peer dependency so that we only have
```json
  "peerDependencies": {
    "graphql": "^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0",
    "graphql-tag": "^2.0.0",
    "mobx": "^4.0.0 || ^5.0.0",
    "mobx-state-tree": "^3.0.0"
  },
```
I guess this means moving createUseQueryHook in react.tsx to a separate package or dynamically require react

mweststrate:
Looking at it, graphql-tag and graphql could actually also removed.
Removing the peer deps doesn't change anything for how the deps are
actually handled, they are just a hint to the package managers that a
warning should be printed if they are absent / incompatible. Requires are
always dynamic, regardless whether you specify dependencies at all or not :)

FredyC:
> Requires are always dynamic, regardless whether you specify dependencies at all or not :)

Sadly it's not true with Webpack which would crash when it sees dependency that's missing. Since everything is bundled into a single dist file, there is no way clear way for optional dependency.

Does that mean we would need runtime plugins as well? It could get messy fairly quickly if we ever decide to support eg. Preact, Vue and whatever.

dpnolte:
Yes, I had made a quick example without add react as dependency: https://codesandbox.io/s/mhf4g
Webpack throws a DependencyNotFoundError.

Although not the cleanest solution, I was thinking of using require.resolve() or System.import and catching it

mweststrate:
I might be missing something, but as host app you still need to install
react and such, it is just that as a package you don't have to specify it.
So afaik webpack should work correctly if `mobx mobx-react react reactdom
mst-gql` are all installed in your project

mweststrate:
Note that with peer deps you also have to do that!

FredyC:
@mweststrate That's not the point :) As long as `react` is referenced in a bundled file, Webpack would fail when you don't have it installed. That's important if you want to only use mobx models, but not the `useQuery`.

dpnolte:
I'm sorry if my comments contributed to the confusion. My use case is to use mst-gql without react. For my use case, it would be great if any react related dependencies would not be mandatory to run mst-gql. That is, I don't have to add react then to my app anymore. That is why in the example I didn't add react etc to the host app. 

If this is out-of-scope, I understand. But if there is something that I can do to make the react dependencies optional in this package, then I'm happy to help :)

mweststrate:
Ah check I get it now. Probably this can be fixed by having react.ts as
separate entry point, or even easier, by passing React as argument to the
hooks factory rather than importing it.

Op wo 29 mei 2019 13:18 schreef dpnolte <notifications@github.com>:

dpnolte:
Cool solution! I'll create a PR for that :)

mweststrate:
Thanks!

mweststrate:
Ok, just merged #16, so it can be added on top of that!

Op wo 29 mei 2019 13:47 schreef dpnolte <notifications@github.com>:

mweststrate:
Released as 0.1.2. Thanks @ematipico & @dpnolte !

PR #20, opened 2019-05-30
fix - example 3 twitter generated reactUtils when it shouldn't

dpnolte:
Twitter example 3 was no longer working. When scaffolding the server models, the reactUtils were generated while it shouldn't. It shouldn't be generated because '--modelsOnly' was passed to the scaffolder. 

The presence of the reactUtils causes a typescript compiler error, which blocks the server start in the Twitter example. The reactUtils couldn't be compiled by typescript because createStoreContext requires a store that is based on MSTGQLStore, which it ain't.

The fix can be found  in generate.js. The other changes are coming from the latest scaffolding output changes.

Thanks,
  Derk


mweststrate:
Good catch!

PR #22, opened 2019-06-03
Add support for server side rendering

### LINKED ISSUE #6: SSR support

mattiasewers:
@mweststrate this is an attempt at adding support for ssr and will close #6

```js
 async function preload() {
    const client = RootStore.create(undefined, {
      gqlHttpClient: createHttpClient('http://localhost:4000/graphql')
    });
    await getDataFromTree(<App client={client} />, client)
    const state = getSnapshot(client)
}
```


src/react.tsx
```diff
@@ -31,7 +47,19 @@ function normalizeQuery<STORE extends typeof MSTGQLStore.Type, DATA>(
 ): Query<DATA> {
   if (typeof query === "function") return query(store)
   if (query instanceof Query) return query
-  return store.query(query, variables, { fetchPolicy, raw })
+  if (typeof window === "undefined") {
+    const queryPromise = store.query<DATA>(query, variables, {
```
FredyC: Looks like this is duplicated in both conditional branches. I would suggest moving `pushPromise` call to be conditional only.

src/react.tsx
```diff
@@ -31,7 +47,19 @@ function normalizeQuery<STORE extends typeof MSTGQLStore.Type, DATA>(
 ): Query<DATA> {
   if (typeof query === "function") return query(store)
   if (query instanceof Query) return query
-  return store.query(query, variables, { fetchPolicy, raw })
+  if (typeof window === "undefined") {
```
FredyC: You sure it's enough like that? I don't know all the possible environments that support SSR, just recall seeing some more elaborate checks in various places.

src/react.tsx
```diff
@@ -1,21 +1,37 @@
...
+declare var ReactNamespace: typeof import("react")
 
 export type QueryLike<STORE, DATA> =
   | ((store: STORE) => Query<DATA>)
   | Query<DATA>
   | string
   | DocumentNode
 
-export function createStoreContext<STORE extends typeof MSTGQLStore.Type>(React: typeof ReactNamespace) {
+export function createStoreContext<STORE extends typeof MSTGQLStore.Type>(
```
FredyC: Perhaps I am mistaken, but it could be more correct like this and you wouldn't need to explicitly type those `__promises` later.
```suggestion
export function createStoreContext<STORE extends Instance<typeof MSTGQLStore>(
```
Note: Instance is imported from MST.

Or it might be better to export type alias from `MSTGQLStore.ts` directly.

mattiasewers:
@FredyC Thank you for reviewing! There are no tests for react atm so did not add any

src/react.tsx
mattiasewers: @FredyC you will have to ask @mweststrate about that, that is how it is typed elsewhere

src/react.tsx
mattiasewers: @FredyC maybe like this `(typeof process !== 'undefined') && (process.release.name === 'node')`?

src/react.tsx
mattiasewers: @FredyC how so? I am concered about this though
```js
  if (typeof query === "function") return query(store)
  if (query instanceof Query) return query
```
those cases are not covered and it seems like they should be

FredyC:
> @FredyC Thank you for reviewing! There are no tests for react atm so did not add any

Yea well, it's always better to start with at least some basic tests instead of waiting for someone else to do it. I am not asking for anything extra elaborate, especially with SSR the snapshot testing is pretty relevant.

mattiasewers:
@FredyC you are more than welcome to write one, I'll happily merge tests. I'm just doing what I can to help because I see potential in this.

src/react.tsx
FredyC: Um, is Node only environment capable of SSR? I honestly do not know.

FredyC:
I don't use SSR and frankly, I am not in need for a project like this as I've established my own way of handling GraphQL and I am happy about it.

I think it's kinda a good practice to supply tests when you are trying to contribute some functionality. I know it's no fun, but consider you would add to the integrity of a whole project in case someone else wants to improve this code. And perhaps you will find answers to your questions in https://github.com/mobxjs/mst-gql/pull/22#discussion_r289992401

src/react.tsx
FredyC: I know, but can you try it? I believe it should work.

mattiasewers:
@FredyC I have discussed this issue with the author and want to take this one step at a time to make the implementation right, my way of doing might not even be the best one. If you do not care and do not want to contribute please do not waste my time

FredyC:
You got that quite opposite, don't you think? I was only trying to help with review and stuff. I don't know why do you expect from me to write tests for something you want to implement. Never heard that one before 😆

src/MSTGQLStore.ts
```diff
@@ -14,7 +14,8 @@ export interface RequestHandler<T = any> {
 
 export const MSTGQLStore = types
   .model("MSTGQLStore", {
-    __queryCache: types.optional(types.map(types.frozen()), {})
+    __queryCache: types.optional(types.map(types.frozen()), {}),
+    __promises: types.optional(types.array(types.frozen()), [])
```
mweststrate: better move `__promises` to `.volatile`, as promises can't be serialized

package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "test": "jest test && cd examples/2-scaffolding && yarn start",
     "watch": "jest test --watch",
-    "build": "microbundle --no-compress --external 'mobx,mobx-react,mobx-state-tree,graphql-tag,graphql,react,graphql-request'"
+    "build": "microbundle --no-compress --external 'mobx,mobx-react,mobx-state-tree,graphql-tag,graphql,react,react-dom,graphql-request'"
```
mweststrate: were is `react-dom` coming from?

mweststrate: ah, I see

src/react.tsx
```diff
@@ -1,21 +1,37 @@
 import { DocumentNode } from "graphql"
+import { renderToStaticMarkup } from "react-dom/server"
```
mweststrate: better pass this function in to getDataFromTree, as done with the other functions, this makes sure that the package doesn't depend accidentally on react

src/react.tsx
mweststrate: It's both the same thing, (I actually don't know what is idiomatic, cc @xaviergonz )

src/react.tsx
mweststrate: I think it would be better to be explicity and pass `ssr: true` as environment config (just like the http client and such)

src/react.tsx
```diff
@@ -31,7 +47,19 @@ function normalizeQuery<STORE extends typeof MSTGQLStore.Type, DATA>(
 ): Query<DATA> {
   if (typeof query === "function") return query(store)
   if (query instanceof Query) return query
-  return store.query(query, variables, { fetchPolicy, raw })
+  if (typeof window === "undefined") {
+    const queryPromise = store.query<DATA>(query, variables, {
+      fetchPolicy: fetchPolicy,
+      raw: raw
+    })
+    store.pushPromise(queryPromise.promise)
```
mweststrate: to not leak memory, I think there should also be an `query.finally(() => { store.unpushPromise(queryPromise.promise) }` (cleanPromises is then no longer needed)

mweststrate:
Thanks for the contribution!

Let's not get into a fight, clearly I didn't leave everything tested, as this project is in a stage. Ideally every new PR increases coverage, but if the infrastructure to easily do so is lacking, help with that would be welcome from anyone.

Before finishing this PR, it should be clear how to use the feature however. So at least a section in the Readme. Including a super simple demo project to the examples (or changing an existing one) would greatly help with enabling others to use the feature, and discuss the final api shape once everything is working correctly.

src/react.tsx
mattiasewers: I'll think we should perhaps move this logic to `Query.ts`?

mattiasewers:
@mweststrate thanks! I'll continue the to work on this through the day!

src/react.tsx
FredyC: I was mainly concerned why is it necessary to explicitly type those `__promises`. With the correct `STORE` type, it should be able to infer it, shouldn't it?

src/react.tsx
mattiasewers: how do I unpush a promise?

mweststrate:
declare it as `.volatile(self => ({ __promises: new Set() }))` and you can
use the normal set api's to add / remove promises efficiently

mattiasewers:
@mweststrate moved logic to the query class, added docs and an example 🔥

examples/5-nextjs/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "build": "tsc",
+    "dev": "next",
+    "start": "next start"
+  },
+  "dependencies": {
+    "apollo-server-express": "^2.6.1",
+    "mobx": "^5.9.4",
+    "mobx-react": "^6.0.3",
+    "mobx-state-tree": "^3.14.0",
+    "mst-gql": "file:./my-package1559643209.tgz",
```
dpnolte: could you please remove this line? mst-gql will be installed via install-mst-gql.sh scripts and yarn throws error when doing yarn install if this file dependency is present

examples/5-nextjs/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "scaffold-test",
+  "scripts": {
+    "install-mst-gql": "bash ../../scripts/install-mst-gql.sh",
+    "scaffold": "../../generator/mst-gql-scaffold.js --roots 'Todo' --excludes 'CacheControlScope,Query,Subscription' --outDir src/models/ --format ts http://localhost:4000/graphql",
+    "build": "tsc",
```
dpnolte: should build also include next build? I'm not really an expert on next though, so you probably know that better than I do. I'm also missing a script to start the server?

examples/5-nextjs/server/schema.js
```diff
@@ -0,0 +1,61 @@
+const fetch = require("isomorphic-fetch")
```
dpnolte: missing isomorphic-fetch as dependency?

examples/5-nextjs/server/index.js
```diff
@@ -0,0 +1,34 @@
+const { ApolloServer } = require('apollo-server-express');
+const { createServer } = require('http');
+const express = require('express');
```
dpnolte: missing express as dependency?

examples/5-nextjs/server/index.js
```diff
@@ -0,0 +1,34 @@
...
+server.applyMiddleware({ app });
+
+const webServer = createServer(app);
+
+const graphqlEndpoint = `http://localhost:${PORT}${server.graphqlPath}`;
+
+app.use(cors());
+app.get(
+  '/',
+  expressPlayground({
```
dpnolte: I've got 'an expressPlayground is not defined' when trying to run this. How could I best start the server?

mattiasewers:
@dpnolte Thank you for your comments! The server example should be much simpler now.

mattiasewers:
@mweststrate @dpnolte can I merge this?

mattiasewers:
@dpnolte Are the changes acceptable to you?

dpnolte:
I've checked it and it works great! Thanks for the contribution and sorry for the delay :)
The only minor thing is that the readme in example 5 is not accurate, but I'll update that myself.
Cheers, Derk

examples/5-nextjs/utils/initModels.ts
```diff
@@ -0,0 +1,24 @@
...
+let store: ModelCreationType<RootStoreType>;
+
+export function initializeStore (isServer: boolean, snapshot = null): ModelCreationType<RootStoreType> {
+  if (isServer) {
+    store = RootStore.create(undefined, {
+      gqlHttpClient: createHttpClient('http://localhost:4000/graphql'),
+      ssr: true
+    });
+  }
+  if (store === null) {
```
fullofcaffeine: Shouldn't this be `store === undefined`? I tried this in a project of mine and it failed because if `store` is not defined, it's not `null`.

zenflow: @fullofcaffeine Yup, you're right. But this example app has been updated and it's no longer an issue. I suggest you refer to the code in `master` (main/default) branch instead: https://github.com/mobxjs/mst-gql/blob/master/examples/5-nextjs/pages/_app.tsx

PR #24, opened 2019-06-06
Fix code example in README

johnrees:
Changed `<Home />` to `<App />` as `Home` had not been imported

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ const rootStore = RootStore.create(undefined, {
 // 4
 ReactDOM.render(
   <StoreContext.Provider value={rootStore}>
-    <Home />
+    <App />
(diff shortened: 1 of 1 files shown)


johnrees:
actually, Home is referenced later in the docs, maybe the import should be changed to Home instead

dpnolte:
Thanks, good catch! I guess that @mweststrate wrote these snippets based on example 1:
- https://github.com/mobxjs/mst-gql/blob/master/examples/1-getting-started/src/app/index.tsx
- https://github.com/mobxjs/mst-gql/blob/master/examples/1-getting-started/src/app/Home.tsx

The app component is used there so lets keep it consistent with that. Probably in the future we can illustrate it better with direct references to the example.

PR #36, opened 2019-06-17
generating typescript input types

elie222:
Addresses:
https://github.com/mobxjs/mst-gql/issues/28

PR #39, opened 2019-06-23
fix - interface types not being generated; duplicate INPUT_OWNER switch entry; cleanup example 5

dpnolte:
This PR fixes a couple of things:
- ModelSelector for interfaces were not being generated. That is why the test failed
  - Solution: `type.kind !== "INTERFACE"` was removed in the types filter
- `printTsType` contained duplicate INPUT_OBJECT switch entry 
  - Solution: I've kept the entry added with #36 
- Example 5 readme & package name were not correct
  - Solution: updated readme and change package name


dpnolte:
@elie222 @mweststrate I'm wondering about the review process and when to merge a PR. What would make sense to you guys? Should every PR be checked by at least one of the maintainers? And feature PRs also by @mweststrate ?

PR #47, opened 2019-07-01
allow multiple queries

mattiasewers:
This would allow for multiple queries like so:

```js
export const GET_TODOS = gql`
  query UserWithTodos {
    user {
      id
      username
      fullname
    }
    todos {
      id
      name
    }
  }
`;
```

This is also breaking change because when using `useQuery` with react you to use the key of the query instead of `data` directly

Also unsure about the types

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -3,7 +3,6 @@ import stringify from "fast-json-stable-stringify"
 import { DocumentNode, print } from "graphql"
 
 import { StoreType } from "./MSTGQLStore"
-import { getFirstValue } from "./utils"
 import { observable, action } from "mobx"
 
 export type CaseHandlers<T, R> = {
@@ -101,7 +100,6 @@ export class Query<T = unknown> implements PromiseLike<T> {
       this.store.__cacheResponse(this.cacheKey, data)
     }
 
-    const value = getFirstValue(data)
     if (this.options.raw) {
       this.loading = false
       this.data = data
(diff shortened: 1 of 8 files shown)


elie222:
If it’s a breaking change, do all the examples still work?
I didn’t fully understand what you meant about the change to hooks. What would the new format look like?

mattiasewers:
@elie222 like so instead
```js
const { data } = useQuery(GET_LAUNCH_DETAILS, {
  fetchPolicy: 'cache-first'
});
```

`data` will now have a key for example `data.launches.id` instead of `data.id`

mweststrate:
I guess it makes sense (so far I didn't get why apollo had that intermediate key, since there always seemed to be one, but this explains). 

N.B. for versioning, pre 1.0.0, increasing `x` in  `0.x.y` can be used to signal a breaking change according to semver

src/Query.ts
```diff
@@ -101,15 +100,17 @@ export class Query<T = unknown> implements PromiseLike<T> {
...
-    const value = getFirstValue(data)
     if (this.options.raw) {
       this.loading = false
       this.data = data
       this.onResolve(this.data!)
     } else {
       try {
         this.loading = false
-        const normalized = this.store.merge(value)
+        const normalized: any = {}
```
elie222: So really not sure how much the typing here matters if at all, but i guess we could at least make it:

```ts
const normalized: { [key: string]: any } = {}
```

mattiasewers:
@mweststrate if everything looks good I'd like to merge this!

mweststrate:
@mattiasewers  LGTM!

Issue #51, opened 2019-07-09
Use `maybeNull` instead of `maybe` for default fields


brobertsUPS:
I read the docs on graphql and the default is every field is nullable by default:

https://graphql.org/learn/best-practices/#nullability

I initially ran into this when trying to query a relation type on an `ItemList` component. The query looks like this:

![Screen Shot 2019-07-08 at 5 25 00 PM](https://user-images.githubusercontent.com/5407471/60850722-6c4a5f80-a1a5-11e9-9aa4-02a4a58f51ad.png)

The generated mobx model looks like this:

![Screen Shot 2019-07-08 at 5 23 28 PM](https://user-images.githubusercontent.com/5407471/60850860-f5619680-a1a5-11e9-8d77-8012907573b4.png)

The issue I ran into is for some items not having a project on them and the type not being accurate when I get `project: null` back on the query:

![Screen Shot 2019-07-08 at 5 24 18 PM](https://user-images.githubusercontent.com/5407471/60850746-8d12b500-a1a5-11e9-8e0a-d3d275344f93.png)

I could be wrong but it feels like the generated models should use `maybeNull` to match with graphql and allow types to be null? When I manually edit the generated model file to use `maybeNull` mst-gql correctly loaded my data into the store.

Happy to do a pull request if this makes sense and wasn't intentional!

elie222:
I think you're right. `maybeNull` is more appropriate. Happy for you to
make a PR with the fix.

brobertsUPS:
Sounds good, I will see if I can create a PR for this

zenflow:
@brobertsUPS @elie222 It seems to make sense to use **both** `maybe` & `maybeNull`:
- `maybe`, so value can be `undefined` when value hasn't been loaded from graphql server yet
- `maybeNull` **only for fields that are nullable (i.e. not explicitly non-null) according to the graphql schema**, so value can be `null` when the value loaded from graphql server is `null`.

Right now the field nullability according to the graphql schema is ignored (assumed to be nullable), and all fields (except `id`) can be `null` (which could either mean that it's actually null or that it's not loaded yet).

kimgronqvist:
@zenflow I was just going to create an issue about this.

My thinking is that we wouldn't use maybeNull at all. All fields gets wrapped in types.maybe, and nullable fields are a union of the type and null.

```js
nonNullableStringProp: types.maybe(types.string);
nullableProp: types.maybe(types.union(types.null, types.string))
```

[...]

PR #52, opened 2019-07-15
change default nulable type fields from maybe to maybeNull

pvpshoot:
I think this can be help
#51


brobertsUPS:
Hey @pvpshoot, thanks for getting to this. It dropped off my radar for a bit. I am less familiar with this code base but I wonder if we need to add tests for this?

elie222:
The examples likely need updating too

elie222:
Which are serving as a sort of test suite right now

pvpshoot:
ok, i will try to add some test cases

pvpshoot:
It was rathe hard, tests don't run, examples don't build according readme. i replaced cases wit type.maybeNull, i think all must be good.
please check @elie222 @brobertsUPS

antondomratchev:
I was having a similar issue as #51. My work around was to overload the property with a maybeNull in the Model where needed. I think this fix will make the models compatible with Graphql's default type of null.
Great PR @pvpshoot

generator/generate.js
```diff
@@ -300,16 +300,16 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
...
           )
         case "OBJECT":
           return wrap(
             handleObjectFieldType(fieldName, fieldType),
             useMaybe,
-            "types.maybe(",
+            "types.maybeNull(",
             ")"
-          )
+
```
chrisdrackett:
```suggestion
)
```

chrisdrackett: missing a closing `)` for `wrap`

chrisdrackett:
@pvpshoot I think you are missing a `)` but other than that this works when run in @brobertsUPS and my project.

Thanks for the PR!

elie222:
Would like to get this one but have some merge conflicts unfortunately

PR #56, opened 2019-07-16
export enum for use in typescript

### LINKED ISSUE #54: export enum types when outputting to typescript
for example instead of:

```
/* This is a mst-gql generated file, don't modify it manually */
/* eslint-disable */
/* tslint:disable */
import { types } from "mobx-state-tree"

/**
* LogType
*/
export const LogTypeEnum = types.enumeration("LogType", [
        "DAILY",
  "MONTHLY",
  "YEARLY",
      ])
```

output:

```
/* This is a mst-gql generated file, don't modify it manually */
/* eslint-disable */
/* tslint:disable */
import { types } from "mobx-state-tree"

export type LogEnumType = "DAILY" | "MONTHLY" | "YEARLY"

/**
* LogType
*/
export const LogTypeEnum = types.enumeration("LogType", [
        "DAILY",
  "MONTHLY",
  "YEARLY",
      ])
```

The name `XXXEnumType` is terrible and too close to the MST name of `XXXTypeEnum`. Any suggestions of a better name?

chrisdrackett:
closes #54

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -31,6 +31,7 @@ function generate(
   const objectTypes = [] // all known OBJECT types for which MST classes are generated
   const inputTypes = [] // all known INPUT_OBJECT types for which MST classes are generated
   const knownTypes = [] // all known types (including enums and such) for which MST classes are generated
+  const enumTypes = [] // enum types to be imported when using typescript
   const toExport = [] // files to be exported from barrel file
   let currentType = "<none>"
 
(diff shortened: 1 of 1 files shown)


chrisdrackett:
using this I think it might be better to export a typescript enum vs. just the values. Will update later today.

chrisdrackett:
ok, I updated this and was able to add the typescript types for enums to the rootStore as well.

chrisdrackett:
@elie222 fixed the merge conflicts here

Issue #49, opened 2019-07-05
Add option to generate base models at different folder


Amareis:
On big schemas where is too much files :)

![image](https://user-images.githubusercontent.com/4806617/60712578-c1ca0680-9f30-11e9-9aca-d0dd266a9303.png)

elie222:
Lol. May be smart for us to group these files by type so each type has its
own folder.

PR #60, opened 2019-07-24
add --separate option to group model files into separate folders

Zyc0017:


Zyc0017:
@elie222 fix issue #49 , please review

PR #64, opened 2019-07-30
add ability to use localStorageMixin in react-native

chrisdrackett:
This PR lets a use provide their own storage method to the `localStorageMixin` in case they are not in a web environment.


elie222:
Published as version 0.3.2 on npm. (But I've been a bit lazy with `CHANGELOG.md` for the last updates :()

chrisdrackett:
@elie222 looks like this didn't make it into `0.3.2`, fyi. I'll try and update.

chrisdrackett:
@elie222 I updated the changelog and added a tag but I can't currently publish to npm (don't have permissions)

PR #69, opened 2019-08-01
Update Examples

chrisdrackett:
As part of this PR I updated example 1 in the following ways:

- upgrade to using babel
- upgrade deps

I've never run the examples myself and wasn't having much luck getting example 1 to run. I switched over to using babel-typescript so that our current typescript errors don't break the build. I got past that part and the page now shows up in the browser, but with errors.

At this point I feel like I might be missing steps, and think my errors probably have something to do with conflicting react versions between the base `node_modules` and the `node_modules` in example 1. (`Uncaught Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`)

I don't have any more time right not to spend on it but figured it can't hurt to put my work so far up.

closes #67

PR #71, opened 2019-08-01
remove outdated instructions about running examples

chrisdrackett:
As far as I can tell it isn't possible to run the examples this way. I still haven't been able to get example 1 running myself (and haven't had a chance to check the others). But it will probably reduce at least some confusion in the short term to remove outdated instructions.


mweststrate:
Closing in favor of #69

PR #77, opened 2019-08-09
fix query types

### LINKED ISSUE #66: types on `data` returned from useQuery not correct
I think this is related to #47. it seems now that data is returned in the following format given something like `queryUser()` I now get:

```
data = {
   user: {
     email: 'test@test.com'
   }
}
```

however data is still typed as an instance of `User` so doing `data.user.email` gives me an error:

`Property 'user' does not exist on type 'ModelInstanceTypeProps…`

seems like we need to update the types for the `data` bit of the `useQuery` hook.

mattiasewers:
This fixes #66

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -8,7 +8,7 @@ import { observable, action } from "mobx"
 export type CaseHandlers<T, R> = {
   loading(): R
   error(error: any): R
-  data(data: T): R
+  data(data: { [key: string]: T }): R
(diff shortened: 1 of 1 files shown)


elie222:
Looks good but we have a conflict right now.

mattiasewers:
@elie222 fixed

elie222:
Circle ci now fails

PR #81, opened 2019-08-13
add correct query hook type

mattiasewers:
fixes the ci build

### DIFF
--- a/src/react.tsx
+++ b/src/react.tsx
@@ -62,7 +62,7 @@ export type UseQueryHookResult<STORE, DATA> = {
   store: STORE
   loading: boolean
   error: any
-  data: DATA | undefined
+  data: { [key: string]: DATA } | undefined
   // prevData: DATA | undefined // set of previously fetched values, in case the query was replaced
   query: Query<DATA> | undefined
   setQuery: (query: QueryLike<STORE, DATA>) => void


veeral-patel:
@mweststrate Hey -- had the same question. How should we iterate over our data with this new query type?

Thanks for the project btw!

beepsoft:
@veeral-patel Well, I just reverted back to `data: DATA | undefined` and it works fine from there on.

mattiasewers:
I cannot replicate this, try upgrading typescript

beepsoft:
@mattiasewers I just checked out the project and built it using the TS version provided by the project (3.5.3) and I get this error:
```
..../examples/1-getting-started/src/app/Home.tsx
Error:(14, 23) TS2339: Property 'map' does not exist on type '{ todos: TodoModelType[]; }'.
Error:(14, 27) TS7006: Parameter 'todo' implicitly has an 'any' type.
```

beepsoft:
Could it be related to this?

https://github.com/microsoft/TypeScript/issues/6041

mattiasewers:
@beepsoft nope, you were right this is a bug. I have rewritten the type implementation again and it should be correct now, I'll have a pr shortly

beepsoft:
@mattiasewers Great, thank you for your effort!

veeral-patel:
@beepsoft Hey -- thanks for your reply! What did you mean by this? Did you change your mst-gql version?

> @veeral-patel Well, I just reverted back to data: DATA | undefined and it works fine from there on.

beepsoft:
@veeral-patel I replaced every occurence of `{[key: string]: DATA}` or` {[key: string]: T}` to `DATA` and `T` respectively in the code (in react.tsx and Query.ts). (In earlier versions of mst-gql that's how these were defined).

But let's wait for @mattiasewers coming up with a better solution!

PR #86, opened 2019-08-30
Add allowFieldArgs

### LINKED ISSUE #46: relations with arguments not showing up on model
I have a schema that looks like the following:

```
type User {
  createdAt: DateTime!
  email: String!
  followers(after: String, before: String, first: Int, last: Int, orderBy: UserOrderByInput, skip: Int, where: UserWhereInput): [User!]
  following(after: String, before: String, first: Int, last: Int, orderBy: UserOrderByInput, skip: Int, where: UserWhereInput): [User!]
  id: ID!
  items(after: String, before: String, first: Int, last: Int, orderBy: ItemOrderByInput, skip: Int, where: ItemWhereInput): [Item!]
  itemsCreated(after: String, before: String, first: Int, last: Int, orderBy: ItemOrderByInput, skip: Int, where: ItemWhereInput): [Item!]
  name: String!
  projects(after: String, before: String, first: Int, last: Int, orderBy: ProjectOrderByInput, skip: Int, where: ProjectWhereInput): [Project!]
  updatedAt: DateTime!
}
```

but after running `mst-gql` I get the following:

```
/**
 * UserBase
 * auto generated base class for the model UserModel.
 */
export const UserModelBase = MSTGQLObject
  .named('User')
  .props({
    __typename: types.optional(types.literal("User"), "User"),
    createdAt: types.maybe(types.frozen()),
    email: types.maybe(types.string),
    id: types.identifier,
    name: types.maybe(types.string),
    updatedAt: types.maybe(types.frozen()),
  })
  .views(self => ({
    get store() {
      return self.__getStore<typeof RootStore.Type>()
    }
  }))

export class UserModelSelector extends QueryBuilder {
  get createdAt() { return this.__attr(`createdAt`) }
  get email() { return this.__attr(`email`) }
  get id() { return this.__attr(`id`) }
  get name() { return this.__attr(`name`) }
  get updatedAt() { return this.__attr(`updatedAt`) }
}
export function selectFromUser() {
  return new UserModelSelector()
}

export const userModelPrimitives = selectFromUser().createdAt.email.name.updatedAt
```

I found that if I remove the args MST picks things up correctly. For example if I change followers to be `followers: [User!]` I then see it correctly in my store:

`followers: types.optional(types.array(MSTGQLRef(types.late((): any => UserModel))), []),`

I don't actually use these args, so I can remove them from the generated `schema.graphql` but I would need to setup a script or similar to do this every time the file is generated.

beepsoft:
Fields with arguments are ignored by default by the generator. With
--allowFieldArgs one can have approriate definitions generated for
such fields as well.

Fixes #46

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -23,7 +23,8 @@ function generate(
   excludes = [],
   generationDate = "a long long time ago...",
   modelsOnly = false,
-  noReact = false
+  noReact = false,
+  allowFieldArgs = false
 ) {
   excludes.push(...buildInExcludes)
 
@@ -294,7 +295,7 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
     let modelProperties = ""
     if (type.fields) {
       modelProperties = type.fields
-        .filter(field => field.args.length === 0)
+        .filter(field => (allowFieldArgs ? true : field.args.length === 0))
         .map(field => handleField(field))
         .join("\n")
     }
--- a/generator/config.js
+++ b/generator/config.js
@@ -1,8 +1,8 @@
-const { existsSync } = require("fs");
-const { resolve } = require("path");
-const cosmiconfig = require('cosmiconfig');
+const { existsSync } = require("fs")
+const { resolve } = require("path")
+const cosmiconfig = require("cosmiconfig")
 
-const explorer = cosmiconfig('mst-gql');
+const explorer = cosmiconfig("mst-gql")
 
 const defaultConfig = {
   excludes: [],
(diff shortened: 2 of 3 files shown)


generator/generate.js
```diff
@@ -294,7 +295,7 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
     let modelProperties = ""
     if (type.fields) {
       modelProperties = type.fields
-        .filter(field => field.args.length === 0)
```
chrisdrackett: I'm curious why this line was here in the first place (@mweststrate) but if things work fine without it I'm all for just removing it entirely vs. adding an `allowFieldArgs` option.

scherler: hmm yeah, I thought maybe orginally it was meant as `field.args.length !== 0`  but does not make sense neither down the line. I agree @chrisdrackett just drop only that line (would as well reduce all the unreleated formating change)  :trollface:

chrisdrackett:
closing due to duplicate #90

PR #89, opened 2019-09-11
Fix query hook data return type again...

mattiasewers:


veeral-patel:
@mattiasewers Thank you for putting this together! Your team seems to move very quickly in response to bug fixes, and I really appreciate that. This project has saved so many lines of boilerplate data fetching code; please keep up the terrific work!

PR #97, opened 2019-09-19
Fix SSR issues #94 & #95 & (somewhat incidentally) bug #91

### LINKED ISSUE #94: [SSR] getDataFromTree ignores (doesn't wait for) queries triggered after loading first queries
Sometimes we have a component that makes a query then renders some component(s) that makes more queries.

Currently `getDataFromTree` does not fulfill those queries that are made after the first query(s) are done. The page is server-rendered while all queries beyond the first "layer" are still not loaded. The second layer of queries is actually triggered [server-side] by the second and final rendering of the react element tree, but the results come after the page has been sent.

You can see this issue in action in my branch for #93 . When page opens in browser, the list(s) are populated, but the "Assignee" for each list item is still "Loading...".

This is a problem when you want to take advantage of SSR while taking the approach of using components (or groups of components) that declare their own data dependencies (e.g. PR #93) rather than the approach of depending on data that was fetched in one monolithic (i.e. for every element on the page) request by code somewhere up the hierarchy of react elements. In this [first, components-requesting-their-own-data] case you would typically have multiple "layers" of queries to make. (Hopefully at some point we will be able to opt-in to batching queries together in each layer (#4), or at least have them deduplicated automatically, which seems straightforward & uncontroversial enough.)

This is something that works in apollo, as you can see from inspecting [it's `getDataFromTree` implementation](https://github.com/apollographql/react-apollo/blob/10de04b246f0404df7fb0416050530b1ef81c5eb/packages/ssr/src/getDataFromTree.ts#L24-L55): `process()` (which includes rendering tree) calls itself recursively while `hasPromises()`.

So in a similar fashion, we should in [our `getDataFromTree` implementation](https://github.com/mobxjs/mst-gql/blob/2808d6ea12f14f7fd2eab87ca2559f3519af82ce/src/react.tsx#L21-L31) probably be, after awaiting promises, re-rendering and checking for more promises. Something like:
```ts
export async function getDataFromTree<STORE extends typeof MSTGQLStore.Type>(
  tree: React.ReactElement<any>,
  client: STORE,
  renderFunction: (
    tree: React.ReactElement<any>
  ) => string = require("react-dom/server").renderToStaticMarkup
): Promise<string> {
  while (true) {
    const html = renderFunction(tree)
    if (client.__promises.size === 0) {
      return html
    }
    await Promise.all(client.__promises)
  }
}
```

But if you are using the default fetchPolicy "cache-and-network" for your queries, that will cause an infinite loop, since each rendering will cause a network request for **every** query, including ones that have already been requested and cached.

Apollo deals with this (as well as the next issue I am about to document) by, depending on `ssrMode` & `ssrForceFetchDelay` options, forcing fetchPolicy to "cache-only" for queries with one of the fetch policies that always uses network.

The semantics of those options are that `ssrMode: true` disables network fetch policies for good, while using `ssrForceFetchDelay: 100` will disable them for 100ms and then start allowing them. I think we could simplify this: 

I propose that when the existing `ssr` [client constructor] option is `true`, we disable network fetch policies (by forcing the policy to "cache-first" as needed) only during server-side rendering and initial client-side re-rendering, and, without needing to opt-in via a `ssrForceFetchDelay`, after that allow use of whatever fetch policy developer wants. 

@mattiasewers @chrisdrackett What do you think? Happy to put in a PR for this!

### LINKED ISSUE #95: [SSR] Queries made on server are repeated on client when page loads
This issue assumes queries use the fetchPolicy of "cache-and-network" (default) or "network-only" . It's not an issue if using "cache-first" or "cache-only". And SSR won't work at all if using "no-cache".

You can see this issue in action in `examples/5-nextjs` (if you work around some minor issues starting it), or in my branch that works on it (#93):

1. server-rendered page loads, showing full set of `todos` and no "loading" indicator
2. once page re-renders (milliseconds later), all the queries for the page are fired again and "loading" indicator is shown
3. once query results are received, "loading" indicator is hidden and (probably) the same data is shown.

This seems inefficient and needless, since the requests that are being repeated were originally made only milliseconds before (not considered stale). Besides performance & bandwidth concerns, it also isn't good that (in repeating the query) it puts the query into the `loading` state again, which might be reflected in our UI.

This can be solved with what I proposed in #94:
> I propose that when the existing `ssr` [client constructor] option is `true`, we disable network fetch policies (by forcing the policy to "cache-first" as needed) only during server-side rendering and initial client-side re-rendering, and, without needing to opt-in via a `ssrForceFetchDelay`, after that allow use of whatever fetch policy developer wants. 

When I say "network fetch policies" I am referring to (as per the context, 2 paragraphs above) "fetch policies that always uses network.". To solve this issue, those need to be disabled during client-side re-rendering. (To solve #94, they also need to be disabled during server-side rendering)

zenflow:
I started working on fixing #94 with my proposed solution (force fetchPolicy to 'cache-first' when on server or first render on client, and render+wait repeatedly in getDataFromTree) but found that getDataFromTree was still entering an infinite loop, since `store.query()` still pushes a promise (to store.__promises) when it's a cache hit (i.e. just getting the results from the cache, synchronously).

So it was necessary to rework the control-flow of the `Query` class, so that we only `store.pushPromise()` when there is some actual async work to do, and otherwise just set the results (`data` or `error`) and the `promise` (which is already settled). The current code in master follows a shared path of execution once it has data, regardless of whether it is from the cache or from a fresh response (and with no use of a param/flag to indicate which). That is also the cause #91. I separated out two execution paths `fetchResults` & `useCachedResults`, and simplified the async control-flow (I have never found it helpful to pass around a promise's `resolve` & `reject` handlers. There is a reason they are only meant to be called inside the promises callback, as in `new Promise(callback)`). Instead of repeating the problematic code in the new `useCachedResults` path, there I put the `data` through a new `store.deflate` method (removes data properties from objects of known types) before putting it through `store.merge` (only if query is not "raw"). This is all done in 00a8366.

I didn't make the 2nd step (see #91) of removing the (above mentioned) data properties of known types from the `__queryCache` (which would be awesome) because the `query` method can be called with the `raw` option, and that will need the full, inflated (un-`deflate`d) data to be in the cache. To deal with that I propose a new option for the `RootStore` constructor: `cacheMode: "raw" | "normalized"` to indicate how you want your data cached. With 'raw' things behave exactly as they do already. With 'normalized', the data is put through `store.deflate()` before saving to __queryCache. The data no longer needs to be `deflate`d just before `merge`ing. If query is called with `raw: true`, the data should be put through a new `store.inflate()` method before being returned, which will add the data props back to objects of known types. This way we can save many bytes in __queryCache (for SSR or serializing/deserializing for localStorage) while keeping first-class support for "raw" queries. 

@chrisdrackett @mattiasewers What do you think of my proposed `cacheMode: "raw" | "normalized"` option? We could default it to "raw" for now, since that is the current behavior, and in the version 1 release switch it to "normalized" since that is (I imagine) what people would want in most cases for performance.

The other commits in this PR are pretty straight-forward I think. 
- 2a8950f Fix SSR, closes #94, closes #95 
- c726b82 Add `noSsr: boolean` query option to skip fetching query on server side
- bf4e35a [5-nextjs] Demo (1) layered SSR queries. (2) noSsr query option

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -107,30 +107,29 @@ export class Query<T = unknown> implements PromiseLike<T> {
     this.loading = true
     this.promise = this.store
       .rawRequest(this.query, this.variables)
-      .then(
-        action((data: any) => {
-          // cache query and response
-          if (this.fetchPolicy !== "no-cache") {
-            this.store.__cacheResponse(this.cacheKey, data)
-          }
-          this.data = this.options.raw ? data : this.store.merge(data)
-          this.loading = false
-          return this.data!
-        })
-      )
-      .catch(
-        action(error => {
-          this.error = error
-          this.loading = false
-          return Promise.reject(error)
-        })
-      )
+      .then((data: any) => {
+        // cache query and response
+        if (this.fetchPolicy !== "no-cache") {
+          this.store.__cacheResponse(this.cacheKey, data)
+        }
+        return this.options.raw ? data : this.store.merge(data)
+      })
     if (this.store.ssr) {
       this.promise = this.promise.finally(() => {
         this.store.unpushPromise(this.promise)
       })
       this.store.pushPromise(this.promise)
     }
+    this.promise.then(
+      action((data: any) => {
+        this.loading = false
+        this.data = data
+      }),
+      action(error => {
+        this.loading = false
+        this.error = error
+      })
+    )
   }
 
   private useCachedResults() {
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -23,6 +23,8 @@ export interface QueryOptions {
   fetchPolicy?: FetchPolicy
 }
 
+const isServer: boolean = typeof window === "undefined"
+
 export class Query<T = unknown> implements PromiseLike<T> {
   @observable loading = false
   @observable.ref data: T | undefined = undefined
@@ -41,7 +43,12 @@ export class Query<T = unknown> implements PromiseLike<T> {
   ) {
     this.query = typeof query === "string" ? query : print(query)
     // possible optimization: merge double in-flight requests
-    this.fetchPolicy = options.fetchPolicy || "cache-and-network"
+    let fetchPolicy = options.fetchPolicy || "cache-and-network"
+    const isSsrConditions = this.store.ssr && (isServer || !store.__afterInit)
+    if (isSsrConditions) {
+      fetchPolicy = "cache-first"
+    }
+    this.fetchPolicy = fetchPolicy
     this.cacheKey = this.query + stringify(variables)
     const inCache = this.store.__queryCache.has(this.cacheKey)
     switch (this.fetchPolicy) {
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -3,7 +3,7 @@ import stringify from "fast-json-stable-stringify"
 import { DocumentNode, print } from "graphql"
 
 import { StoreType } from "./MSTGQLStore"
-import { observable, action } from "mobx"
+import { action, observable } from "mobx"
 
 export type CaseHandlers<T, R> = {
   loading(): R
@@ -32,8 +32,6 @@ export class Query<T = unknown> implements PromiseLike<T> {
   public promise!: Promise<T>
   private fetchPolicy: FetchPolicy
   private cacheKey: string
-  private onResolve!: (data: T) => void
-  private onReject!: (error: any) => void
 
   constructor(
     public store: StoreType,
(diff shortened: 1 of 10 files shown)


zenflow:
Btw, I believe this PR is ready for review and merging. This fixes the essential problem of #91 & two other issues and we can add the new `cacheMode` option in a follow-up PR.

src/Query.ts
```diff
@@ -21,8 +21,11 @@ export type FetchPolicy =
 export interface QueryOptions {
   raw?: boolean // If set, the response data is returned verbatim, rather than parsing them into the relevant MST models
   fetchPolicy?: FetchPolicy
+  noSsr?: boolean
 }
 
+const isServer: boolean = typeof window === "undefined"
```
chrisdrackett: Might this come back as true for `react-native`?

zenflow: I'm not sure, but there's a good chance yes. But why would you use `ssr: true` for a react-native app? I believe with a react-native app you would have no place to use (and no use for) SSR.

Out of curiosity I spent a bit of time researching the correct way to define this, but couldn't find out anything certain of whether it would work with react-native. But I think it should not matter here.

chrisdrackett: true! I made this comment before reading the below, so I think we're good :)

zenflow: Another option is `const isServer = !process.browser` (and is the way I like to do it, and I even use it inline) but that depends on using a module bundler that [shims the global `process` object](https://www.npmjs.com/package/process) (webpack & browserify do it by default) and also I got a [type error when trying to use it in typescript](https://github.com/zeit/next.js/issues/2177)

chrisdrackett:
@zenflow this looks good for me and appears to work in the examples, so that's good!

That being said, some things are a bit out of my wheelhouse. I'm wondering if someone else can review this as well before we merge. @elie222? @mattiasewers?

as for the `cacheMode` are their downsides (other than doing more testing) to just defaulting to `normalized` vs. adding an option? Is it that we need to add a `inflate` functionality first?

zenflow:
> as for the cacheMode are their downsides (other than doing more testing) to just defaulting to normalized vs. adding an option?

The downside of normalizing by default is making "raw" queries just a tiny bit not as nice:
- Extra work first normalizing & converting it to MST objects, then de-normalizing & converting to plain js objects. Could be noticeable for large sets of data.
- Possibly receiving extra data props that weren't requested. Good & fine when we are working with our model instances, but seems unexpected & maybe undesired when working with `raw` requests.

I don't actually have much of a use for "raw" queries **or** a "raw" cache mode right now, but I think we should fully preserve it since I could see it being useful in some situations, maybe debugging, to completely circumvent all the crazy things we're doing in our models.

zenflow:
> I don't actually have much of a use for "raw" queries or a "raw" cache mode right now, but I think we should fully preserve it since I could see it being useful in some situations, maybe debugging, to completely circumvent all the crazy things we're doing in our models.

Actually, I can't really imagine any uses for "fully raw" queries at all.. even for debugging because I can see the network response in chrome devtools.. and obviously even that would not have been enough to make the optimization.

When someone passes `raw: true` for a query, they just want to get back js objects instead of model instances, for whatever reason. It's really pretty simple. The data can come from the same place as with non-raw queries.

zenflow:
@chrisdrackett I must agree now that there's no need for adding an option, and we can just always keep a normalized cache. 

I can add the `inflate` functionality easily (following the pattern of `merge` & `deflate`), and make the changes to keep cache normalized. 

I will do that soon and commit it in this PR if it hasn't already been reviewed & merged. See https://github.com/mobxjs/mst-gql/pull/97#issuecomment-533235623

zenflow:
Made one major fixup: bc7da9b Fixup 00a836, in deflate(), only remove props of *root* type objects. If we removed props of objects of types that are known but not root (as my code was doing before), then we would never get them back (in most cases, i.e. unless you've saved the instance in the store yourself) because `merge()` only adds instances of root types to the store.

zenflow:
00a836 is a mistake. It should be 00a8366.

mattiasewers:
@zenflow Good job! I've gone through everything and nothing seems to break tests, types or examples.

@chrisdrackett This can be merged whenever you see fit

zenflow:
@chrisdrackett do you think this could be merged? (I don't care when it's released) I prefer to do the normalizing __queryCache bit in a follow-up PR if that's ok. Also I think it still needs a bit of discussion first, see issue #99 which I just opened to carry on the discussion.

chrisdrackett:
Yeah, lets get it merged and I'll work on getting a release cut later this week!

PR #100, opened 2019-09-24
[optimization] Deduplicate identical concurrent query requests

zenflow:
As per @mweststrate's comment at https://github.com/zenflow/mst-gql/blob/4f37e7444775aeafb9b883bb2e17933937109607/src/Query.ts#L46
> // possible optimization: merge double in-flight requests

This is an important optimization when rendering a bunch of components that declare their own data dependencies, since dependencies are commonly repeated. For example, in examples/5-nextjs, in the "Todos" list, where todos are assigned to either Rick or Morty, when loading we want to request data for each of those `User`s a maximum of once. With the current code in master, when loading we have one user request per todo and therefore some duplicate requests.

I changed the public API of `RootStore` by removing `pushPromise` & `unpushPromise` and replacing them with a private `__pushPromise`. I think it makes sense to be private, and I think it makes sense for it (`__pushPromise`) to handle unpushing the promise after it settles, since it's something that a caller will always need to happen. Let me know if you disagree with either statement.

/cc @chrisdrackett @mattiasewers

### DIFF
--- a/src/MSTGQLStore.ts
+++ b/src/MSTGQLStore.ts
@@ -19,7 +19,7 @@ export const MSTGQLStore = types
   })
   .volatile((self): {
     ssr: boolean
-    __promises: Set<Promise<unknown>>
+    __promises: Map<string, Promise<unknown>>
     __afterInit: boolean
   } => {
     const {
@@ -29,7 +29,7 @@ export const MSTGQLStore = types
     } = getEnv(self)
     return {
       ssr,
-      __promises: new Set(),
+      __promises: new Map(),
       __afterInit: false
     }
   })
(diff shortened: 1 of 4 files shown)


zenflow:
Pushed another commit: (7d653e0) "Fixup 77678ec, fix tests". I can't really figure out why the test was passing before, but the test is now correct and passing. Immediately after firing the query, we should expect 1 promise, not 0 promises. We should expect 0 promises only after the query settles.

chrisdrackett:
looks good to me! We have one conflict atm, but other than that I don't think anything else is blocking merge I can think of

PR #103, opened 2019-09-25
Drop "raw" option for queries & keep __queryCache normalized (#99)

### LINKED ISSUE #99: [optimization] Keep __queryCache normalized
First brought this up in issue #91. I'm separating it out since it's a related, but essentially different, issue.
 
> Filter/remove [now-] unused data from __queryCache, so root type objects in query results [in __queryCache] have no properties except __typename and id (the two that are needed to find the object among our cached model instances).

See PR #97 for some discussion about how this should work with "raw" queries.

Based on my branch for #97 I started implementing this went the approach of keeping __queryCache normalized by default (no option), and to not worry about "raw" queries too much as long as they return the results as plain js objects. See https://github.com/zenflow/mst-gql/compare/fix-ssr...zenflow:normalize-cache.

The problem with that is that (aside from returning some extra data that was not requested in the query selector as mentioned in #97, which is unavoidable unless we filter based on the parsed query selector) many schemas have recursion (e.g. `user.todos[0].user.todos[0].user`), so the result of a "raw" query may be an object of "circular structure" which can't be serialized to JSON.

image

With that approach, the raw query results are really not the same type of object as they were before, and we are really getting far away from the notion of "raw" results that haven't been changed/enhanced.

At this point I propose we drop the "raw" option for queries, and focus on making mst-gql work perfectly when we want query results cached/enhanced/deduplicated/etc (which. if I'm not mistaken, is the vast majority of the time). I don't think anyone will miss the "raw" option. What do you guys think? If nobody can think of a potential use-case for it, perhaps we can remove it and see if anyone misses it and if they really needed it.

/cc @chrisdrackett @mattiasewers

zenflow:
This is how I would like to resolve #99 (follow-up to issue #91 / PR #97).

Notice that none of the examples needed to be changed 😃 

I added a new snapshot test before making the changes, so you can observe the effect of this optimization via the changes to the snapshot in 34cb585

/cc @chrisdrackett @mattiasewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ async function preload() {
 
 # 🍿 In-depth store semantics 🍿
 
-mst-gql generates model types for every object type in your graphql definition. (Except for those excluded using the `excludes` flag). For any query or mutation that is executed by the store, the returned data will be automatically, and recursively parsed into those generated MST models (unless the `raw` flag is set). This means that for any query, you get a 'rich' object back. Finding the right model type is done based on the GraphQL meta field `__typename`, so make sure to include it in your graphql queries!
+mst-gql generates model types for every object type in your graphql definition. (Except for those excluded using the `excludes` flag). For any query or mutation that is executed by the store, the returned data will be automatically, and recursively parsed into those generated MST models. This means that for any query, you get a 'rich' object back. Finding the right model type is done based on the GraphQL meta field `__typename`, so make sure to include it in your graphql queries!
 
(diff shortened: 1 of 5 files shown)


mattiasewers:
@zenflow @chrisdrackett I am ok with dropping `raw` I think anyone who'd want to use this package want to use mst 😄 

There is a small conflict in the readme now.

zenflow:
> There is a small conflict in the readme now.

Resolved

PR #106, opened 2019-09-25
Support customizing model base

zenflow:
The last change on my wish list:

This is a small but important change to the generator to "Generate & consume customizable module containing ModelBase" (26b887b).

The next time a user runs scaffold, a `ModelBase` module **which can be manually edited** is placed in outDir:

```ts
import { MSTGQLObject } from "mst-gql"

export const ModelBase = MSTGQLObject
```

and `xxxModel.base` module is changed to import & use that `ModelBase` instead of `MSTGQLObject` directly.

This will allow us (mst-gql users/consumers) to extend & enhance the common ancestor of every model, similar to how we can already extend & enhance the RootStore & the individual models. This means we have full customization control of mst classes, and can do a lot of neat things on top of what mst-gql provides! (and ideally move the good ideas of general usefulness into this repo)

@chrisdrackett @mattiasewers If this is cool with you then I will finish this WIP by updating the README & examples 1 through 4.


mattiasewers:
@zenflow @chrisdrackett This change is ok with me, if we could get another ok from the other contributors I can merge this

chrisdrackett:
I think this looks really good as well! Lets get some documentation in this PR and I'm more than happy to merge it!

zenflow:
> Lets get some documentation in this PR and I'm more than happy to merge it!

@chrisdrackett documentation updated! 

I guess we can leave the remaining examples as they are, since the migration to having a `ModelBase` module is seamless. So only the 5-nextjs example integrates this change so far.

chrisdrackett:
probably should have gotten this in with `0.5.0` 😩

zenflow:
We can release 0.6.0 now as well if we want. :)

Are you a user of gitter or spectrum.chat or slack or anything like that for collaboration?

chrisdrackett:
yeah, but we don't currently have anything setup for this project. I'll set something up!

PR #109, opened 2019-09-25
have format command run on entire project.

chrisdrackett:
restablish baseline formatting

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,11 +1,28 @@
+# config
+.DS_Store
+.gitignore
+.prettierignore
+
 # build files
 .circleci/
 .rts2*/
 .vscode/
 dist/
+.cache/
+.next/
+
+# test files
+__snapshots__
 
 # mst-gql generated files
 **/models/index.*
 **/models/reactUtils.*
 **/models/*.base.*
 **/models/*Enum.*
+
+# yarn
+yarn.lock
+yarn-error.log
+
+# etc
+LICENSE
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,15 +1,8 @@
-# config
-.DS_Store
-.gitignore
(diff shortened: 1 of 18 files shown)


.prettierignore
```diff
@@ -1,11 +1,28 @@
+# config
+.DS_Store
+.gitignore
+.prettierignore
```
zenflow: @chrisdrackett Can we replace many of these lines with a single `.*` line to ignore all dot-files & dot-folders? That would also cover IDE-related folders like the `.idea` folder Jetbrains WebStorm creates. Currently prettier is picking up some files in that folder and complains that the file type is unsupported.

package.json
```diff
@@ -34,7 +34,7 @@
     "test": "jest test && cd examples/2-scaffolding && yarn && yarn start",
     "watch": "jest test --watch",
     "build": "microbundle --no-compress --external mobx,mobx-react,mobx-state-tree,graphql-tag,graphql,react,react-dom,graphql-request",
-    "format": "prettier --write ./**/*"
+    "format": "prettier --write '**/*'"
```
zenflow: This line!

So removing the `./` bit magically fixed the problems I was having running prettier on some directories.

The single-quotes around `**/*` isn't Windows compatible because cmd.exe doesn't strip quotes from around command-line arguments. But I know the quotes are desired in linux so that the glob pattern is passed as-is to prettier command, and not expanded by the shell into megabytes of filenames from node_modules. So for now I'll just patch my local working copy here.

chrisdrackett: yeah, for some reason without the quotes prettier ignores `examples` for example. I have no idea why...

zenflow:
@chrisdrackett I almost forgot to thank you for fixing this up when I needed it! Thanks!

PR #112, opened 2019-09-30
Control Import Order

RXminuS:
The nature of GraphQL is that it is very easy to end up with circular dependencies. Simply something as

```graphql
type Node {
    edges: [Edge!]!
}

type Edge {
    from: Node!
    to: Node!
}
```

can quickly become problematic. mstGQL currently doesn't allow any elegant way of controlling the import order resulting in some funky behaviour. In my opinion I don't think any of the `.base` files should be separate and would prefer one giant `base.generated.ts` file as it would avoid a lot of the problems but also make it easier with Typescript types on the resolvers etc. But also realise that it probably would be an unpopular move.

So instead I followed Westrate's [excellent blogpost](https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de) about use of `internal.ts` barrel files.

This change simply makes sure that all the modules now use `import {...} from "./internal"` rather than individual files. You can then use something like [barrelsby](https://www.npmjs.com/package/barrelsby) to generate an internal.ts file that simply exports every module. You can then move around items in that file to control the loading order (Enums at the top etc).

This code isn't well tested yet so probably needs a bit of scrutiny before merging, but it works for our repo.

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -41,22 +41,12 @@ function generate(
 
   const interfaceAndUnionTypes = resolveInterfaceAndUnionTypes(types)
 
-  generateModelBase()
   generateTypes()
   generateRootStore()
   if (!modelsOnly && !noReact) {
     generateReactUtils()
   }
   generateBarrelFile(files)
-
-  function generateModelBase() {
-    const entryFile = `\
-import { MSTGQLObject } from "mst-gql"
-
-export const ModelBase = MSTGQLObject
-`
-    generateFile("ModelBase", entryFile)
-  }
 
   function generateTypes() {
     types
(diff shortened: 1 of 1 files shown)


RXminuS:
Oh, realise that I might have copied in the edits as I made them in my code editor which was still on v0.4.x but pasted them into a v0.5.x fork.

Might be easiest to just cherry-pick them over instead, basically all I did was change the generation of multiple import statements to a single import statement for "./internal" that imports all the modules from there.

Also we might want to add the generation of the "./internal" file the first time so that it works out of the box? Or maybe we want to hide this behaviour behind a flag instead?

zenflow:
@RXminuS 

I'm sorry, I don't really understand the problem this is trying to solve, so I can't give much of a review, but anyways.

Some changes merged to master after the 0.5 release are reverted by this.. see changes to `generate.js` in https://github.com/mobxjs/mst-gql/pull/106/commits/26b887b6a070b7ea4c8df5c1b4764365c297f417

RXminuS:
Yeah I will redo it on the latest master branch. I will also see if I can add an example to show the problem, but it’s basically that SomeModelBase is being imported circularly before its defined. Which happens because imports are currently happening directly from the modules

PR #114, opened 2019-09-30
Work on support for abstract (union or interface) types

zenflow:
## Primary fix

[generate] Use MSTGQLRef /w root types also for abstract type fields (b2f3760)

Currently in master, the generator uses MSTGQLRef around root types only when that root type is appearing as a simple/concrete field type. When the root type is appearing as a part of an abstract (interface or union) field type, MSTGQLRef is not used.

MSTGQLRef *always* needs to be used with root types, because the mst nodes of root types are already living in their dedicated prop on the root store. We will otherwise get errors like `[mobx-state-tree] Cannot add an object to a state tree if it is already part of the same or another state tree. Tried to assign an object to '/owner', but it lives already at '/users/y'` when loading query data.

*(edited)* For a concrete example, see `tests/lib/abstractTypes` where the `Repo` type has an `owner` field with the (abstract) interface type `Owner` which is implemented by (concrete) types `User` and `Organization`...

this:

https://github.com/mobxjs/mst-gql/blob/ad4682613f9d11c4e64323f302432ebc17894273/tests/lib/abstractTypes/models/RepoModel.base.js#L21

becomes:

https://github.com/mobxjs/mst-gql/blob/4e9c1ee8f31363553a297c14ee9f1e47ac1270ae/tests/lib/abstractTypes/models/RepoModel.base.js#L21

Note: You can see the tests failing with the aforementioned error at commit 47ecbe2 `[tests/lib/abstractTypes] Add User & Organization to root types`.

## Other changes 

Fixed defined root types in `tests/lib/abstractTypes`. Removed SearchResult (4ffa1f4) since it doesn't have an ID and thus doesn't qualify as a root type. Added User & Organization (47ecbe2), since they can qualify as a root type, and that is probably what the user would want it to be

I noticed we are preventing object fields from being a reference to another object of the same type, which can't be right, so I fixed that with commit c1a41e4 `[generate] Allow object field to be *reference* to object's own type`.

I also noticed that, similar to the main issue I was fixing, we are forgetting to do something for abstract field types that we are doing for simple/concrete field types. With commit da66cb1 `[generate] Use frozen for unknown types also for abstract type fields` we consistently use `types.frozen()` for unknown types.

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -377,7 +377,9 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
       const isSelf = fieldType.name === currentType
 
       // this type is not going to be handled by mst-gql, store as frozen
-      if (!knownTypes.includes(fieldType.name)) return `types.frozen()`
+      if (!knownTypes.includes(fieldType.name)) {
+        return `types.frozen()`
+      }
 
       // import the type
       const modelType = fieldType.name + "Model"
@@ -410,6 +412,10 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
 
       const interfaceOrUnionType = interfaceAndUnionTypes.get(fieldType.name)
       const mstUnionArgs = interfaceOrUnionType.ofTypes.map(t => {
+        if (!knownTypes.includes(t.name)) {
+          return `types.frozen()`
+        }
+
         // Note that members of a union type need to be concrete object types;
         // you can't create a union type out of interfaces or other unions.
         const subTypeClassName = t.name + "Model"
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -638,12 +638,16 @@ ${rootTypes
         if (returnType.kind === "OBJECT" && excludes.includes(returnType.name))
           return ""
         // TODO: probably we will need to support input object types soon
-        if (returnType.kind !== "OBJECT") {
+        const isReturnTypeAnObject =
+          returnType.kind === "OBJECT" ||
+          returnType.kind === "UNION" ||
+          returnType.kind === "INTERFACE"
+        // TODO: for now, we only generate queries for those queries that return objects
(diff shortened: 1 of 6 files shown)


zenflow:
Note that stylistic changes are due to 5f24785 where we stop formatting the generated XXXModelSelector modules with prettier

zenflow:
#### Made an additional notable change, improving support for abstract (i.e. union & interface) types:

0e19a2f `Generate methods for queries having abstract return types` (& b6b7ea1 `Add test for query methods having abstract return types`)

So if you have a schema like:

```graphql
type Query {
  list: [SomeAbstractType]
}
```

the generated (base) root store will have a query method for "list" (i.e. `store.queryList()`) as expected.

RXminuS:
@zenflow Is this ready for review/merge? Is there anything I can help with?

zenflow:
@RXminuS Yup, it's ready. I have been using a custom version of mst-gql that has this change, and I haven't found any corrections to make, or any more obvious with issues using abstract types. If you want to give it a review, that would be great!

fullofcaffeine:
Hi! Is this ready to be used? I'll probably need to introduce abstract types in my schema soon and it'd be superb to have it supported out-of-the-box in `mst-gql` :)

zenflow:
I think this will be easier to land if I break it into a few more-focused PRs. Hopefully I can get some review from @RXminuS @fullofcaffeine  😺 

1. Fix and enhance tests for abstract types (`tests/lib/abstractTypes`)
2. Always use `MSTGQLRef` when defining fields having a root type. Currently this is done when that type is a concrete type, but not if it's an abstract type.
3. Always generate a method for a query. Currently this is done when the query return type is concrete, but not if it is abstract.
4. Always use `types.frozen()` for unknown types. Currently this is done for fields of concrete types, but not for fields of abstract types.

PR #120, opened 2019-10-03
replace finally to retain old react-native support

### LINKED ISSUE #118: remove promise.finally to fix react-native
promise.finally can cause issues in some versions of react-native. I see we recently added a new instance of it here:

https://github.com/mobxjs/mst-gql/blob/5479a544b7c2e9dc2e0e83b2ebad393839318947/src/MSTGQLStore.ts#L144

I'm not super strong in how promises work atm, so I'm wondering if someone else could take a look.

/cc @zenflow

chrisdrackett:
closes #118

### DIFF
--- a/src/MSTGQLStore.ts
+++ b/src/MSTGQLStore.ts
@@ -141,9 +141,8 @@ export const MSTGQLStore = types
       rawRequest,
       __pushPromise(promise: Promise<{}>, queryKey: string) {
         self.__promises.set(queryKey, promise)
-        promise.finally(() => {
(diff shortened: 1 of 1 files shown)


RXminuS:
Maybe just put a comment before it to reference this issue?

zenflow:
> Maybe just put a comment before it to reference this issue?

I wouldn't. Then issue #118 would become an actual burden on the project. Avoiding use of `finally` would become an actual constraint. And we would want someone to come and undo it after expo upgrades it's version of RN.

Issue #124, opened 2019-10-17
add ability to skip merge on query/mutation


chrisdrackett:
I'm thinking more about the optimistic update of new items case and how I might handle that when using `most-gql`. 

Building off @elie222's suggestion here: https://github.com/mobxjs/mst-gql/issues/61#issuecomment-515704098

I'm wondering if it's useful to add the ability to turn off the merge step for a mutation. Then you could do something like:

```js
const query = store.mutateAddTodo({ title: 'new todo' }, undefined, () => {
  const todo = Todo.create({
    id: 'temp-id',
    __typename: 'Todo',
    title
  })
}, {
  skipMerge: true
})

query.case({
  data: (data) => {
    todo = {…todo, …data}
  }
})
```

so in the above pseudocode you could run the mutation with a (new) `skipMerge` option. This way when the mutation is complete you can merge the updated data (id and anything else calculated on the server) in with the item that was created as part of the optimistic update.

This avoids the current potential problem of having 2 of the same item to juggle on the client. In this case the item created by the optimistic update stays around and is only updated based on the results from the server.

chrisdrackett:
https://github.com/mobxjs/mobx-state-tree/issues/887 would be nice :)

chrisdrackett:
my gut is that without https://github.com/mobxjs/mobx-state-tree/issues/887 or similar it won't be possible to handle something like this without the following flow:

- add temp item
- await server response
- delete temp item as item from server is added.

This does create a lot of complications for an application (like ours) that is trying to support offline operation. If the user has mutated the temp item or even worse added sub-items to it (say a comment on a post) it creates complications when trying to sync with the server.

fullofcaffeine:
> This does create a lot of complications for an application (like ours) that is trying to support offline operation. If the user has mutated the temp item or even worse added sub-items to it (say a comment on a post) it creates complications when trying to sync with the server.

@chrisdrackett Did you end up using any workarounds for that limitation?

chrisdrackett:
@fullofcaffeine not at the moment. We have this on the roadmap on our project but it isn't something we're currently working on. The solution might require some updates to mst-gql, but as for this specific fix, in light of the current state of MST I'm not sure it is worth keeping this specific issue open.

PR #127, opened 2019-10-18
[WIP] Allow skipping automatic merging of mutation results

brobertsUPS:
This seems like it should be able to solve https://github.com/mobxjs/mst-gql/issues/124

I added `mutationOptions` to allow `skipMerge` which keeps the results of a mutation from being automatically merged into the root store data. This should allow optimistic updates at the project level.

### DIFF
--- a/examples/3-twitter-clone/src/app/models/RootStore.ts
+++ b/examples/3-twitter-clone/src/app/models/RootStore.ts
@@ -1,7 +1,7 @@
 import { Instance } from "mobx-state-tree"
 import { RootStoreBase } from "./RootStore.base"
 import { types } from "mobx-state-tree"
-import { MessageModel } from "./MessageModel"
+import { MessageModel, MessageModelType } from "./MessageModel"
 import { selectFromMessage } from "./MessageModel.base"
 
 export interface RootStoreType extends Instance<typeof RootStore.Type> {}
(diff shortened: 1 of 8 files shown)


examples/3-twitter-clone/src/app/models/RootStore.ts
```diff
@@ -44,7 +44,28 @@ export const RootStore = RootStoreBase.props({
   }))
   .actions(self => ({
     sendTweet(text: string, replyTo = undefined) {
-      return self.mutatePostTweet({ text, user: self.me.id, replyTo })
+      const query = self.mutatePostTweet(
+        { text, user: self.me.id, replyTo },
+        undefined,
+        () => {
+          const tempMessage = MessageModel.create({
+            id: "tempid",
```
chrisdrackett: should we generate a random ID here to make sure things don't overlap?

examples/3-twitter-clone/src/app/models/RootStore.ts
```diff
@@ -44,7 +44,28 @@ export const RootStore = RootStoreBase.props({
...
+            replyTo
+          })
+        },
+        { skipMerge: true }
+      )
+
+      // query.case({
+      //   error: error => console.log(error),
+      //   loading: () => console.log("loading"),
+      //   data: data => console.log(data)
```
chrisdrackett: update the above `MessageModel` here

brobertsUPS:
@chrisdrackett I think this can be closed. After working through more `mobx-state-tree`, I think middleware would allow intercepting these updates and skipping anything that wasn't desired. What do you think?

PR #140, opened 2019-11-09
Feature/fix circular refs with typed refs

godness84:
Another attempt to solve definitively the circular refs issue with typescript.

How it works:
- referenced models are typed as `any` (in this way we break circularity, but we lose typings).
- a `withTypedRefs` function gives the correct typings to the refs.

Compared to #128 it has some pros:
- no need of the `as()` wrapper inside actions.
- no need to split the definition of the model into two parts (with/without refs).
- `withTypedRefs` can be used easily whenever new circular refs are introduced by writing custom model code.
- line changes in `generate.js` file are much less and better written


beepsoft:
@godness84 Should `yarn test` work?

I get:

```
Test Suites: 1 failed, 2 passed, 3 total
Tests:       3 failed, 5 passed, 8 total
Snapshots:   3 failed, 5 passed, 8 total
```

godness84:
@beepsoft only snapshots are failing, I have not updated them yet. 😄

beepsoft:
@godness84 Yep, that's what I've seen as well. 🤓 Otherwise your solution seems to be working well without any TS errors and with proper typing, wheee!

Do you plan to port the other examples to use the models you generate?

The only problem I have so far is not related to your code: IntelliJ seems to be really slow finding out the types when I have a huge graphql schema (and with Hasura it is _always_ huge) and many MST models generated for it. VSCode is much faster but it still needs some time to figure out everything when I press `.+control+space`.

chrisdrackett:
@godness84 I'm running this on the examples and in example 3 I'm seeing that in the following code:

```js
export const RootStore = RootStoreBase.actions(self => ({
  log() {
  }
}))
```

`RootStoreBase` is typed as any. I see the same in `RootStore.base.ts`. I don't see the same in anything else.

Other than that this works for me in my circular example (what was once example 6). Our project is a bit of a mess at the moment as we're in the middle of a huge migration. If we can get all the examples and tests updated and squared away I think this will be good to go!

chrisdrackett:
I'm not sure if this is related to this PR, but I've been attempting to use this in our real product and I'm running into the following on some of the prop definitions:

```
Type 'IOptionalIType<ISimpleType<"Item">, [undefined]>' is not assignable to type 'string | number | boolean | Date | IAnyType'.
  Property '[$type]' is missing in type 'IOptionalIType<ISimpleType<"Item">, [undefined]>' but required in type 'IAnyType'.
```

I'm still investigating what might be causing this, but because I see it in the base files I figured I could post here.

godness84:
I suppose your custom code introduced new circular refs somewhere, because
when there are circular refs, strange things happen and TS goes crazy.

Make sure that when one model has a ref to others, it uses "any" and the
"withTypedRefs" trick.

When necessary, make explicit the return type of your function when it
depends on the type returned by other model's functions.

Il mer 13 nov 2019, 23:23 Chris Drackett <notifications@github.com> ha
scritto:

chrisdrackett:
ok, I'll look around. I'm actually running into this with a fresh scaffold (so no added anything). I'll see if I can reduce it down to a simpler failure case

chrisdrackett:
ok, so if I just take my schema and render it into a blank project everything is fine. I guess my problem is either how I'm linking mst-gql into my project or something unrelated to this PR.

beepsoft:
There's a bounty for solving this issue, so @godness84 you should really go for it 🙂

https://www.bountysource.com/issues/76246469-circular-dependencies-between-models

chrisdrackett:
@godness84 this looks great! I'm going to merge and release this now!

chrisdrackett:
@godness84 I'm adding this into my project and want to use the experience to update the docs.

Given the following extension to the root store do I need to do anything?

```js
export const RootStore = RootStoreBase
  .props({
    newProject: types.maybe(ProjectModel),
  })
```

what I did is:

```js
type Refs = {
  newProject: ProjectModelType
}

export const RootStore = withTypedRefs<Refs>()(RootStoreBase
  .props({
    newProject: types.maybe(ProjectModel as any),
  })
)
```

this seems to work, but just wanted to verify it is correct!

godness84:
@chrisdrackett oh great!!! 🎉 

What you did in `RootStore` is correct.

About the `markComplete()` action, I'm not sure that it's required to declare the type of `item`. Sometimes TS goes crazy exactly as you described and by restarting it (I launch the `TypeScript: Restart TS server` command in VSCode) it stops showing the error.

However, if declaring the type of `item` works, it means that it's the right solution 😅

examples/3-twitter-clone/src/app/models/MessageModel.ts
```diff
@@ -38,7 +38,7 @@ export const MessageModel = MessageModelBase.views(self => ({
     },
     loadReplies() {
       if (!loadReplyQuery) {
-        loadReplyQuery = self.store.loadMessages(0, 100, self.id)
+        loadReplyQuery = self.store.loadMessages("", 100, self.id)
```
godness84: `loadMessages` expects a string as first argument instead of a number. So I needed to change it. Don't know how it could compile correctly before 🤔 and mostly I don't know if I changed it correctly. However, I tried the demo and it works.

godness84:
@chrisdrackett uh oh, I forgot to mention one thing about this PR.
I commented the line, take a look. Sorry for pointing it out late

examples/3-twitter-clone/src/app/models/MessageModel.ts
godness84: in `examples/3-twitter-clone/src/app/models/RootStore.ts` at line `54`, `""` is passed to `loadMessages()` so I assume I fixed it correctly

RXminuS:
@godness84 bounty should now be payed out! Congrats ;-)

godness84:
@RXminuS thank you very much! 🎉 😄

PR #150, opened 2019-11-21
update how optional types are generated

### LINKED ISSUE #149: update types for optional mutation variables
Right now a mutation variable that is optional will output in the root store as 

```js
isComplete: boolean | undefined
```

however as far as I know in typescript this isn't correct. it should be:

```js
isComplete?: boolean
```

otherwise if you omit `isComplete` typescript will give an error:

```Property 'isComplete' is missing in type …```

we should update this to use the other format.

I believe the code that needs to change lives here: https://github.com/mobxjs/mst-gql/blame/master/generator/generate.js#L700

### PR DESCRIPTION
**Pull request by chrisdrackett**
closes #149

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -568,7 +568,7 @@ ${ifTS(
     .map(
       t =>
         `\nexport type ${t.name} = {\n${t.inputFields
-          .map(field => `  ${field.name}: ${printTsType(field.type)}`)
+          .map(field => `  ${printTsType(field)}`)
           .join("\n")}\n}`
     )
     .join("")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by zenflow** [comment]
Will have a look on Sunday

File: generator/generate.js [hunk 1]
```diff
@@ -733,26 +731,39 @@ ${optPrefix("\n    // ", sanitizeComment(description))}
...
-  function printTsType(type, isRoot = true) {
+  function printTsType(field, name, canBeUndefined = true) {
+    let typeValue
+    let type
+
+    if (!name) {
+      name = field.name
+      type = field.type
+    } else {
+      type = field
```
brobertsUPS: In the above three lines, can `field.type` and `field` be the same shape? i.e. can `field` and `field.type` have the same properties? Why do we need them both?

chrisdrackett: yes, based on the way a field is described it could be single level or nested (in the case of being `NON_NULL`)

brobertsUPS: 👍

PR #155, opened 2019-12-08
Add --namingConvention option to generator

beepsoft:
Currently only one option is supported: js.

--namingConvention js will generate type names in PascalCase, other names in
camelCase. This can be especially useful in case of Hasura generated schemas
where all names default to snake_case.

The js naming convention also generates different root collection names: these
will be camelCased and properly pluralized. Without --namingConvention js the
root  names are all lowercase and and a plural "s" is appended.
The +"s" however is not always valid English plural in case the root type's
name is  eg. "EmptyBox" as it will become "emptyboxs".
With  --namingConvention js it will be "emptyBoxes".

If --namingConvention is not set or set to anything else than "js" then the
default names provided by the graphql schema will be used.

MSTGQLStore.ts and mergeHelper.ts have also been updated so that root collection
name generation uses the same transformation as at generation time. For this
to work configureStoreMixin() receives the optional namingConvention value as
the third parameter and store.getCollectionName() is used at runtime to convert
__typename to a root collection name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,7 @@ The `mst-gql` command currently accepts the following arguments:
 - `--modelsOnly` Generates only models, but no queries or graphQL capabilities. This is great for backend usage, or if you want to create your own root store
 - `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
+- `--namingConvention asis|js` When set to `js` it will generate type names and root collection names to conform to Javascript naming conventions, ie. type names will be PascalCased, and root collection names camelCased. The default (`asis`) value generates type names matching the original types names provided by the graphql schema
 - `source` The last argument is the location at which to find the graphQL definitions. This can be
   - a graphql endpoint, like `http://host/graphql`
   - a graphql files, like `schema.graphql`
--- a/README.md
+++ b/README.md
@@ -416,7 +416,7 @@ The `mst-gql` command currently accepts the following arguments:
 - `--modelsOnly` Generates only models, but no queries or graphQL capabilities. This is great for backend usage, or if you want to create your own root store
 - `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
-- `--namingConvention asis|js` When set to `js` it will generate type names and root collection names to conform to Javascript naming conventions, ie. type names will be PascalCased, and root collection names camelCased. The default (`asis`) value generates type names matching the original types names provided by the graphql schema
+- `--dontRenameModels` By default generates model names from graphql schema types that are idiomatic Javascript/Typescript names, ie. type names will be PascalCased and root collection names camelCased. With `--dontRenameModels` the original names - as provided by the graphql schema - will be used for generating models.
 - `source` The last argument is the location at which to find the graphQL definitions. This can be
   - a graphql endpoint, like `http://host/graphql`
   - a graphql files, like `schema.graphql`
--- a/examples/1-getting-started/src/app/components/Todo.tsx
+++ b/examples/1-getting-started/src/app/components/Todo.tsx
@@ -1,10 +1,10 @@
 import * as React from "react"
 import { observer } from "mobx-react-lite"
 
-import { TodoModelType } from "../models"
+import { Todo } from "../models"
 import { useQuery } from "../models/reactUtils"
 
-export const Todo = observer(({ todo }: { todo: TodoModelType }) => {
+export const Todo = observer(({ todo }: { todo: Todo }) => {
   const { setQuery, loading, error } = useQuery(undefined)
 
   return (
--- a/examples/1-getting-started/src/app/components/Todo.tsx
+++ b/examples/1-getting-started/src/app/components/Todo.tsx
@@ -4,7 +4,7 @@ import { observer } from "mobx-react-lite"
 import { TodoModelType } from "../models"
 import { useQuery } from "../models/reactUtils"
 
-export const Todo = observer(({ todo }: { todo: Todo }) => {
+export const Todo = observer(({ todo }: { todo: TodoModelType }) => {
   const { setQuery, loading, error } = useQuery(undefined)
 
   return (
--- a/examples/1-getting-started/src/app/components/Todo.tsx
+++ b/examples/1-getting-started/src/app/components/Todo.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import { observer } from "mobx-react-lite"
 
-import { Todo } from "../models"
+import { TodoModelType } from "../models"
 import { useQuery } from "../models/reactUtils"
 
 export const Todo = observer(({ todo }: { todo: Todo }) => {
--- a/generator/config.js
+++ b/generator/config.js
@@ -13,7 +13,7 @@ const defaultConfig = {
   outDir: "src/models",
   roots: [],
   noReact: false,
-  namingConvention: "asis" // supported option: "js"
+  namingConvention: "js" // supported option: "js", "asis"
 }
 
 exports.getConfig = function getConfig() {
@@ -40,6 +40,8 @@ exports.mergeConfigs = function mergeConfigs(args, config) {
     modelsOnly: !!args["--modelsOnly"] || config.modelsOnly,
     forceAll: !!args["--force"] || config.force,
     noReact: !!args["--noReact"] || config.noReact,
-    namingConvention: args["--namingConvention"] || config.namingConvention
+    namingConvention: args["--dontRenameModels"]
+      ? "asis"
+      : config.namingConvention
   }
 }
--- a/generator/config.js
+++ b/generator/config.js
@@ -12,7 +12,8 @@ const defaultConfig = {
   modelsOnly: false,
   outDir: "src/models",
   roots: [],
-  noReact: false
+  noReact: false,
+  namingConvention: "asis" // supported option: "js"
 }
 
 exports.getConfig = function getConfig() {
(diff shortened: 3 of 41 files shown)


chrisdrackett:
@beepsoft any reason not to just creates some sane defaults that transform things like you suggested in your issue and avoid the addition of an option?

beepsoft:
Making these the defaults is fine with me, I just thought that it would be a breaking change and that would effect others already using the package. I also though there maybe other naming conventions people may want to use.

I also suggested some more adjustments to the generated model in this issue: https://github.com/mobxjs/mst-gql/issues/157. Can you please take a look at that too?

beepsoft:
Happy new year to you @chrisdrackett! How should we proceed with this?

I updated my implementation a bit so that I apply the name transformations to the graphql model directly, this way the generator's original code gets much cleaner.

Should I add all the suggestions from #157 and get rid of the --namingConvention option? Is it OK to introduce a breaking change with this?

chrisdrackett:
I’m out until next week, but my gut is still to avoid new options where possible. We are still pre-1.0 so I’m ok with big changes that are improvements to the generated code :)

zenflow:
@beepsoft What is wrong with `snake_case`? Why wouldn't we want to use the exact tokens from the graphql schema in our js code? What benefit does a different *style* of naming offer? 

To me it seems like it would be less complex to just use the names as-is, maintainers of a codebase using this, and especially for maintainers of this package.

zenflow:
@beepsoft Nevermind.. that should be discussed in #157 which I'm reading now

README.md
```diff
@@ -416,6 +416,7 @@ The `mst-gql` command currently accepts the following arguments:
 - `--modelsOnly` Generates only models, but no queries or graphQL capabilities. This is great for backend usage, or if you want to create your own root store
 - `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
+- `--namingConvention asis|js` When set to `js` it will generate type names and root collection names to conform to Javascript naming conventions, ie. type names will be PascalCased, and root collection names camelCased. The default (`asis`) value generates type names matching the original types names provided by the graphql schema
```
chrisdrackett: what if instead of `--namingConvention js` we added a `--dontRenameModels` or similar option. This would make the default idiomatic while still having the option of using the values from the schema if wanted.

generator/config.js
```diff
@@ -12,7 +12,8 @@ const defaultConfig = {
   modelsOnly: false,
   outDir: "src/models",
   roots: [],
-  noReact: false
+  noReact: false,
+  namingConvention: "asis" // supported option: "js"
```
chrisdrackett: My opinion would be to default this to `js`

README.md
beepsoft: That's OK with me. I'll add it soon.

generator/config.js
beepsoft: I just pushed my changes with --dontRenameModels and defaulting namingConvention to "js".

chrisdrackett:
@beepsoft this looks good to me! do you mind updating the examples on your branch? After that I think we're good to merge!

beepsoft:
I updated the examples. Actually, just rerun them and everything seems to work fine with the new naming convention.

I think there should be some words in the README about the breaking changes in the generated code before publishing the new version of the package.

chrisdrackett:
totally agree! 

I'm currently not seeing the examples fully upgraded. For example in 1 it seems like only the root store got updated, so it is trying to import `Todo` (which is still actually `TodoModelType`)

chrisdrackett:
I wonder if as we make the changes to the examples we could write out an entry on "breaking changes" in the changelog so when users are upgrading its as painless as possible.

beepsoft:
> I'm currently not seeing the examples fully upgraded. For example in 1 it seems like only the root store got updated, so it is trying to import `Todo` (which is still actually `TodoModelType`)

Oh, you are right. The non-base types don't get updated by the generator unless --force is used. I'll fix it right now.

beepsoft:
Now I fixed them all, I hope.

> I wonder if as we make the changes to the examples we could write out an entry on "breaking changes" in the changelog so when users are upgrading its as painless as possible.

Sure, it should go to CHANGELOG.md!

beepsoft:
@chrisdrackett Do you need anything else from me to be able to merge?

chrisdrackett:
@beepsoft more time in the day 😅 I was planning on doing the upgrade on my repo and use the process to draft the changelog entry, I just haven't had time yet. Let me try and get to it today

generator/generate.js
```diff
@@ -17,28 +19,42 @@ const buildInExcludes = [
...
+  let origRootTypes = []
 
   const header = `/* This is a mst-gql generated file, don't modify it manually */
 /* eslint-disable */${format === "ts" ? "\n/* tslint:disable */" : ""}`
   const importPostFix = format === "mjs" ? ".mjs" : ""
 
+  // For a model called TodoModel the TS type would originally be called TodoModelType.
+  // Now we make it just "Todo".  But I have this variable so that in case we want to
+  // make it configurable we can just set this back to "ModelType" or to something else.
+  const modelTypePostfix = ""
```
chrisdrackett:
```suggestion
  const modelTypePostfix = "ModelType"
```

generator/generate.js
```diff
@@ -17,28 +19,42 @@ const buildInExcludes = [
...
   const enumTypes = [] // enum types to be imported when using typescript
   const toExport = [] // files to be exported from barrel file
   let currentType = "<none>"
+  let origRootTypes = []
 
   const header = `/* This is a mst-gql generated file, don't modify it manually */
 /* eslint-disable */${format === "ts" ? "\n/* tslint:disable */" : ""}`
   const importPostFix = format === "mjs" ? ".mjs" : ""
 
+  // For a model called TodoModel the TS type would originally be called TodoModelType.
```
chrisdrackett:
```suggestion
  // For a model called `TodoModel` the TS type would be called `TodoModel` by default.
```

generator/generate.js
```diff
@@ -17,28 +19,42 @@ const buildInExcludes = [
...
+  // Now we make it just "Todo".  But I have this variable so that in case we want to
```
chrisdrackett:
```suggestion
  // `modelTypePostfix` is added to the end of this to create a unique typename and avoid conflicts.
```

generator/generate.js
```diff
@@ -17,28 +19,42 @@ const buildInExcludes = [
...
+  // make it configurable we can just set this back to "ModelType" or to something else.
```
chrisdrackett:
```suggestion
  // so `modelTypePostfix` as `ModelType` will give you a final type of `TodoModelType` for the model `Todo`
```

beepsoft:
Well, for me the `TodoModelType` is very artificial sounding when in fact this _is_ a Todo object that we want to store and work with. Of course, the current mst-gql examples were written with `TodoModelType` in mind and therefore the react component's name now clash. But if you were to start a new Todo example you would still have this problem. How would you call your Todo data record and how would you call the component, which displays this record? `Todo` vs `TodoItem`? Or `TodoObject` vs `Todo`?

I would also be glad to hear what @mweststrate thinks about it.

I myself already started using this naming schema in one of my projects (actually refactored the project to use the non-ModelType ending names) and it is much easier to read and understand this way.

chrisdrackett:
I think there might be a discussion around the model names ‘Todo’ vs. ‘TodoModel’ that we could have outside this issue. For now we use the model name ‘TodoModel’ so I feel the type ‘TodoModelType’ is appropriate. We could have that discussion here as well, but it might hold up the rest of the PR :)

mweststrate:
The Model was there, to leave more wiggle room for people to customize things, e.g. create an own `Todo` type that extends `TodoModel`, or have React components that are called `Todo`, so not having models with the same names is nice. Especially if you get a dump of generated things in your application code, it is nice that the 'nice names' aren't take already by that what is generated, and I kinda like the partial signal of `...Model`: this is managed by mst-gql as well. But all just my 2 cents :) Feel free to take any other direction, I don't really have a stake in the game.

beepsoft:
Thanks @mweststrate! I was also thinking about it and since the `...ModelType` names are generated into the non-base model files that are - hopefully - generated only once and where users can add customizations - actions, views, etc. -, one such customization can be the renaming of the automatically generated name (from `TodoModelType` to `Todo`).

So, @chrisdrackett, I agree, let's use the `...ModelType` postfixed names for now and maybe later there could be a customization option for it if needed.

beepsoft:
@chrisdrackett, you are probably busy with other stuff, but let me know if you need anything from me to be able to merge this PR.

chrisdrackett:
I think we just need to update the code and examples to go back from `Todo` to `TodoModelType` and we should be good to go! I might have time this week, but It’ll probably be Friday

chrisdrackett:
@beepsoft sorry for not having time for this yet! Just wanted to let you know it hasn't slipped my mind, I've just been super busy. Let me know if you can get to it and I'm happy to press the merge button :D

beepsoft:
@chrisdrackett Sure, I will do the re-renaming of `Todo` to `TodoModelType`, etc. This is all that's needed, right?

beepsoft:
@chrisdrackett I reverted everything so that for a TodoModel the type name is again TodoModelType.

examples/1-getting-started/src/app/components/Todo.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from "react"
 import { observer } from "mobx-react-lite"
 
-import { TodoModelType } from "../models"
+import { Todo } from "../models"
```
chrisdrackett:
```suggestion
import { TodoModelType } from "../models"
```

examples/1-getting-started/src/app/components/Todo.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { useQuery } from "../models/reactUtils"
 
-export const Todo = observer(({ todo }: { todo: TodoModelType }) => {
+export const Todo = observer(({ todo }: { todo: Todo }) => {
```
chrisdrackett:
```suggestion
export const Todo = observer(({ todo }: { todo: TodoModelType }) => {
```

Issue #141, opened 2019-11-18
update `localStorageMixin` so that it is possible to clearTimeout


chrisdrackett:
We've run into a case where a user logs out of our app and we clear local storage only to have `localStorageMixin` repopulate it shortly after due to the timeout. I wonder if there is a way within mst-gql to avoid this case as part of the mixin so a user won't have to worry about it on their end.

preetb123:
`applySnapshot(self, {})` resets the cache. Put it inside a separate action and call it after logout

rajeevAgilehub:
> `applySnapshot(self, {})` resets the cache. Put it inside a separate action and call it after logout

thanks, it worked..
One more question "can we reset a specific data, not all ?"

chrisdrackett:
@rajeevAgilehub you could do this just by updating the object sent into `applySnapshot` to contain only the data from the current store you want to keep.

PR #166, opened 2020-02-14
Local storage control

special-character:
This PR aims to solve this issue: https://github.com/mobxjs/mst-gql/issues/141

It is possible for the local storage of an app to be saved after the user logs out which can cause problems. I have added two actions to the `localStorageMixin` which can control when snapshots can be saved to localStorage.

In this example, I add something to the cart and logout before the throttle timeout. Then, after the app is logged out, the local storage is set again. This allows me to refresh the page without logging in and still has all of my local storage data.

Local Storage After Logout

This is a simple PR but I am not sure if there is a more reactive way to do this.

### DIFF
--- a/examples/4-apollo-tutorial/client/src/models/RootStore.js
+++ b/examples/4-apollo-tutorial/client/src/models/RootStore.js
@@ -61,6 +61,7 @@ export const RootStore = RootStoreBase.props({
   loginStatus: loginStatus,
   cartItems: types.array(types.string)
 })
+  .extend(localStorageMixin())
   .views(self => ({
     get me() {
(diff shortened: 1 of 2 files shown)


special-character:
The more I think about this, it might be better to just add an action to clear local storage that applies an empty snapshot to self.

special-character:
Closing this in favor of doing `applySnapshot(store, {})` in this issue: https://github.com/mobxjs/mst-gql/issues/141

special-character:
Closing this in favor of doing `applySnapshot(store, {})` in this issue: https://github.com/mobxjs/mst-gql/issues/141

PR #194, opened 2020-03-30
Task/#9 mandatory fields

Matth10:
Depends on #9 

Now it is possible to pass some mandatory fields like so `--mandatoryField 'User.name, etc'`. This will change the mobx-state-tree type's definition of all the fields passed in.

#### Example
Not Mandatory:
`name: types.union(types.undefined, types.number)`
Mandatory:
`name: types.string`

- [x] update doc
- [x] use it in one example


chrisdrackett:
@Matth10 can we talk about the use case for this option? It's still possible to query without these fields and I'm guessing the result will cause errors in MST. I want to understand the drivers before I review! :)

Matth10:
@chrisdrackett I think the goal of this option is to set some fields that need to be queried when we query a specific Model. 

For example when we fetch some fields on the User we always need to ask for the name for example.

chrisdrackett:
right, but this code does not actually control what is queried, just the MST model. Should this code also rewrite or edit queries to always include those fields? If not, aren't errors possible if you use the wrong query?

Matth10:
> However, in strict typescript, this will require a lot of non-null assertions.
Also, it can be very good to make it mandatory that queries for a certain type also fetch a certain field.

Yes I should include the field in the query if it's a mandatory one

chrisdrackett:
@Matth10 this was closed by mistake due to branch renaming. If you want to keep working on this please feel free to reopen!

PR #195, opened 2020-03-30
remove experimental warning

### LINKED ISSUE #193: Get rid of warning in README
I've been looking for something like this library for a LONG time, it's basically a dream come true. 

But I almost decided against using it because of the (extremely effective!) warning in the README about the lack of maintainers. Of course, as it turns out there are now half a dozen maintainers, including one whose company is using this in production and is (presumably) motivated to keep the wheels from falling off this lib 😛

I think if we get rid of that warning this lib will start getting the love it deserves 🤙

chrisdrackett:
closes #193


rdewolff:
So are we production ready now?

chrisdrackett:
Not sure how to define that, but I would use this library in production as it stands. Most the heavily lifting post-generation is MST so as long as you trust that you should be good to go!

Issue #224, opened 2020-05-06
Local Storage Mixin Saves Potentially Stale Data


special-character:
We were working with our application and noticed that local storage wasn't saving some data after the app loaded. I looked into the `localStorageMixin` and added two `onSnapshot` listeners directly next to where the local storage is set by `onSnapshot` ((here](https://github.com/mobxjs/mst-gql/blob/master/src/localStorageMixin.ts#L39))

```
              onSnapshot(self, throttle(function (data) {
                console.log('W/ THROTTLE SNAPSHOT', data)
              }, throttleInterval))

              onSnapshot(self, function (data) {
                console.log('W/O THROTTLE SNAPSHOT', data)
              })
```

```
W/O THROTTLE SNAPSHOT 
{
  "ck9u8b1xs0000dSetunl8kb9d": Object {
    "__typename": "Project",
    "id": "ck9u8b1xs0000dSetunl8kb9d",
    "title": "!!!",
  },
  "ck9u8caxg00003h6011s8zpaz": Object {
    "__typename": "Project",
    "id": "ck9u8caxg00003h6011s8zpaz",
    "title": "@@@",
  },
}

W/ THROTTLE SNAPSHOT                                                                                            
{
  "ck9u8b1xs0000dSetunl8kb9d": Object {
    "__typename": "Project",
    "id": "ck9u8b1xs0000dSetunl8kb9d",
    "title": "!!!",
  },
}
```

After looking at the source code, the `throttle` function schedules a call to the snapshot function with the arguments of the first call. i.e. if there are multiple calls to `onSnapshot` within the `throttleInterval`, the local storage mixin will call the `onSnapshot` callback with the first snapshot and lose any updates that happen after.

I will work on a PR for this.

PR #225, opened 2020-05-06
Clear timeout and schedule a new throttle with the correct args

special-character:
closes: https://github.com/mobxjs/mst-gql/issues/224

This updates the `throttle` function in `localStorageMixin` to save with correct args. The function clears the timeout that was scheduled and re-schedules it with the latest args and the new time.

I mostly took this code from https://github.com/github/mini-throttle.

Here is an example of the two side by side and how they work:
https://codesandbox.io/s/upbeat-williamson-djyqb?file=/src/App.tsx

### DIFF
--- a/examples/3-twitter-clone/src/app/models/RootStore.ts
+++ b/examples/3-twitter-clone/src/app/models/RootStore.ts
@@ -4,6 +4,7 @@ import { types } from "mobx-state-tree"
 import { MessageModel } from "./MessageModel"
 import { selectFromMessage } from "./MessageModel.base"
 import { UserModelType } from "./UserModel"
+import { localStorageMixin } from "mst-gql"
 
 export interface RootStoreType extends Instance<typeof RootStore.Type> {}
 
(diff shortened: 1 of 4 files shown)


dpnolte:
First of all, thanks for the great example and work :)

This looks great, but I would actually be in favour of using a 3rd party package for throttling. This would help keep mst-gql lean and we would benefit from the maturity and testing that is done with packages like mini-throttle and lodash throttle. Could we not simply use mini-throttle?

special-character:
@dpnolte I think we could use a package out of the box but I didn't b/c I was on the fence for a few reasons. 

1) This throttle is used in one place and it is really clear how it works
2) There was actually a bug in min-throttle that I had to fix
3) It was already a custom throttle and I remember reading from Michel (somewhere) about why this was done. I am having a hard time finding it now so this point might be moot

I am happy to add one like `lodash.throttle` but I was on the fence if it was overkill. If there are strong opinion on this I can just add the `lodash.throttle` package

dpnolte:
Thanks for the update!

I did a quick check. And, unfortunately, the uncompressed gzip bundle size increases by ~ 4 kb when including lodash. I guess some of the functionality of lodash leaks into the bundle through dependencies. 

With `throttle-debounce`, the size increases less than 1 kb. I think that is more acceptable for a throttle function. I've added throttle-debounce to your example and seems to work as expected:
https://codesandbox.io/s/sweet-tu-z89uc?fontsize=14&hidenavigation=1&theme=dark

Would this resolve your issue? If so, would you mind updating it to throttle-debounce?

@chrisdrackett What do you think? And do you know why we don't compress the distribution bundle? It would reduce the overall bundle size by 50%

chrisdrackett:
I don't, I haven't really looked at the bundle stuff at all so its basically what was here before I started working on the project. We should take a look at that :D

special-character:
@dpnolte I used lodash b/c it was already in the node_modules from another dep and it is a pretty standard one. I can use `throttle-debounce` instead. The diff between 4kb and 1kb feels arbitrary to me especially if we don't compress things (could be 2kb-1kb) but I am not dogmatic about the change so I will update.

special-character:
@dpnolte `throttle-debounce` works for me. I am not very dogmatic about what lib we use.

I added lodash b/c it is pretty standard and is importing like `lodash/throttle` only imports the throttle function. The larger size is due to it being larger in the source and not from pulling in extra deps. I went ahead and used `throttle-debounce` but the diff between 4kb and 1kb feels arbitrary, especially if it isn't compressed yet.

Thanks for looking at this!

PR #233, opened 2020-05-19
Base action types

special-character:
- Adding types for the base graphql actions so they can be used to dynamically call them.

i.e. I want to call these actions from the RootStore with their string name like:

```
self[graphqlActionName](......args and stuff)
```

This currently complains about TS and doesn't have autocomplete for the possible action names I can call. This PR aims to add types so that it is possible to type these actions.

### DIFF
--- a/examples/4-apollo-tutorial/client/src/models/RootStore.base.js
+++ b/examples/4-apollo-tutorial/client/src/models/RootStore.base.js
@@ -15,14 +15,18 @@ import { UserModel } from "./UserModel"
 import { userModelPrimitives, UserModelSelector } from "./UserModel.base"
 
 
+
+
+
+
 /**
 * Store, managing, among others, all the objects received through graphQL
 */
 export const RootStoreBase = MSTGQLStore
   .named("RootStore")
-  .extend(configureStoreMixin([['LaunchConnection', () => LaunchConnectionModel], ['Launch', () => LaunchModel], ['Mission', () => MissionModel], ['Rocket', () => RocketModel], ['User', () => UserModel]], ['Launch', 'Rocket', 'User']))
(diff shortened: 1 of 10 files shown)


examples/4-apollo-tutorial/client/src/models/RootStore.base.js
```diff
@@ -15,14 +15,18 @@ import { UserModel } from "./UserModel"
...
 /**
 * Store, managing, among others, all the objects received through graphQL
 */
 export const RootStoreBase = MSTGQLStore
   .named("RootStore")
-  .extend(configureStoreMixin([['LaunchConnection', () => LaunchConnectionModel], ['Launch', () => LaunchModel], ['Mission', () => MissionModel], ['Rocket', () => RocketModel], ['User', () => UserModel]], ['Launch', 'Rocket', 'User']))
+  .extend(configureStoreMixin([['LaunchConnection', () => LaunchConnectionModel], ['Launch', () => LaunchModel], ['Mission', () => MissionModel], ['Rocket', () => RocketModel], ['User', () => UserModel]], ['Launch', 'Rocket', 'User'], "js"))
   .props({
-    launchs: types.optional(types.map(types.late(() => LaunchModel)), {}),
+    launches: types.optional(types.map(types.late(() => LaunchModel)), {}),
```
special-character, line 29: I don't this example has been ran in awhile. The `launches` and `"js"` changes above did not come from the new code generation.

examples/5-nextjs/src/models/RootStore.base.ts
```diff
@@ -17,12 +17,29 @@ type Refs = {
...
+mutateUsers="mutateUsers"
+}
+
 /**
 * Store, managing, among others, all the objects received through graphQL
 */
 export const RootStoreBase = withTypedRefs<Refs>()(MSTGQLStore
   .named("RootStore")
-  .extend(configureStoreMixin([['Todo', () => TodoModel], ['User', () => UserModel]], ['Todo', 'User']))
+  .extend(configureStoreMixin([['Todo', () => TodoModel], ['User', () => UserModel]], ['Todo', 'User'], "js"))
```
special-character, line 42: Ditto for this change. I don't think this was updated the last time generation code was updated

PR #235, opened 2020-05-19
Exclude fields from type generation that we exclude in action generation

special-character:
- Don't create types for actions that are skipped in generation

I ran all of the examples and tests (no snapshot changes)

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -666,6 +666,33 @@ ${rootTypes
   }
 
   /**
+   * Returns if this field should be skipped in generation. Can happen if:
+   * 1) The field is in the excludes
+   * 2) The field has a return type that is not supported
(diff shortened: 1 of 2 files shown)


generator/generate.js
```diff
@@ -665,6 +665,33 @@ ${rootTypes
     generateFile("RootStore.base", modelFile, true)
   }
 
+  /**
+   * Returns if this field should be skipped in generation. Can happen if:
+   * 1) The field is in the excludes
+   * 2) The field has a return type that is not supported
+   * @param {*} field from an array of queries or mutations
+   */
+  function shouldSkipField(field) {
```
special-character, line 674: The code in this block was pulled from `generateQueryHelper`

Issue #201, opened 2020-04-09
add at least one enum to an example


chrisdrackett:
Would be good for testing outside our tests.

beepsoft:
I just tried 0.10.0 and actually there is something wrong with enum generation: both the model and the typescript type has the same name. Eg. for an enum `StatusEnum` there should be a file generated with name `StatusEnum.ts` with a typescript type called `StatusEnum` and an mst model `StatusEnumModel`, however the model is now also called `StatusEnum`.

I will look into this now.

beepsoft:
@chrisdrackett It seems that you changed something around this in the a222426e964d8f646774720ec04a2b1e851e8027 and 11e79aa548281874f437eaff2b103da9dd174499 commits. My version generated the `...Enum` and `...EnumModel` names but now you generate the same name for both. Can you take a look at this?

chrisdrackett:
Yeah, enums are not models, I had missed this during the review as none of the mst-gql examples have enums, but found it when updating our project. That being said, maybe I missed something? I don’t see any duplicate names in our enums, but there might be a case I’m not aware of.

beepsoft:
Take a look at `tests/generator/generate.test.js`. There you test for this result: a typescript `enum InterestEnum` and an mst model/type `const InterestEnum`. Here we don't compile these with `tsc` but if we did it would fail because of the name clash.

```javascript
test("enums ending in Enum doesn't have an extra Enum postfix with namingConvention=js", () => {
 ...
  expect(interestEnumFile).toBeTruthy()
  expect(
    hasFileContentExact(interestEnumFile, "export enum InterestEnum {")
  ).toBeTruthy()
  expect(
    hasFileContentExact(
      interestEnumFile,
      'export const InterestEnum = types.enumeration("InterestEnum"'
    )
  ).toBeTruthy()
```

The simplest solution would be to give the mst type some new name. Maybe `InterestEnumType`?

chrisdrackett:
ah, I see, we don't run into this in our app because our enums don't have `enum` in the name. So we end up with something like `export enum ItemType` and `export const ItemTypeEnum` in our `ItemTypeEnum` file. I think the suggestion of adding `Type` to the end of the type is probably the best thing to do here as it matches with what we do with models.

beepsoft:
@chrisdrackett should I create a PR for fixing the above issue?

chrisdrackett:
sure, sorry, we've been extremely busy so I haven't had much time to put to this.

[...]

PR #237, opened 2020-05-22
Add Model suffix to MST enum model to avoid clash with TS enum name

beepsoft:
Also the file names for enums will now end with Model, which makes naming
of files the same for all types.

This is to fix enum issue in #201

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -165,7 +165,7 @@ export const ModelBase = MSTGQLObject
   function handleEnumType(type) {
     const name = type.name
     const enumPostfix = !name.toLowerCase().endsWith("enum") ? "Enum" : ""
-    toExport.push(name + enumPostfix)
+    toExport.push(name + enumPostfix + "Model")
 
     const tsType =
       format === "ts"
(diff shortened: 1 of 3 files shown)


chrisdrackett:
I still feel pretty strongly that we should not be adding `Model` to these names because they aren't models. I think what we should do is:

ItemTypeEnum.ts
```
export enum Role {
   A="A",
   B="B"
}

export const RoleEnumType = types.enumeration("ItemType", [
  "A",
  "B",
])
```

This will mean that if you have a enum with a name like `RoleEnum` you'll end up with `RoleEnumEnumType` for your MST type. For us (and maybe everyone) this type is only used in the generated model so I don't think this is such a big problem.

What do you think?

beepsoft:
OK, I see your point. Here's the new PR: #238.

Did you really mean `RoleEnumEnumType` or just `RoleEnumType`? #238 generates `RoleEnumType` for either a graphql type name is `Role` or `RoleEnum`.

chrisdrackett:
Isn't the fact that we generate differently based on the string the reason we ran into this in the first place? Feels like we're trying to be too smart.

beepsoft:
Well, my original proposal was to get rid of the possible `EnumEnum` like generated names, because many times the enums are named `SomeEnum` in the graphql schema and it is strange to use `EnumEnum` names in code. So, yes we are a bit smart here and in my original implementation I used `Model` suffix for the MST types I just forgot about it. 

But now we can have `Type` suffix and for me it seems that all requirements are satisfied: no double `EnumEnum`s, enum type names are suffixed with `Type` and no clash with TS names.

PR #240, opened 2020-05-23
don't import `RootStore` in `reactUtils` anymore

chrisdrackett:


special-character:
@chrisdrackett I think we need to update snapshots and we should be off to the races

PR #243, opened 2020-06-06
Add optional filter option to localStorageMixin

reekris:
Allows the user to filter which data to store in local storage

- Uses `lodash/pick` since it allows to use dot notation for filtering nested data which is nice. But maybe there's a better alternative?

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,13 +633,14 @@ For examples, see the sections [Loading and rendering your first data](#loading-
 
 ## `localStorageMixin`
 
-The `localStorageMixin` can be used to automatically save the full state of the `RootStore`. By default the store is saved after every change, but throttle to be saved once per 5 seconds. (The reason for the throttling is that, although snapshotting is cheap, serializing a a snapshot to a string is expensive).
+The `localStorageMixin` can be used to automatically save the full state of the `RootStore`. By default the store is saved after every change, but throttle to be saved once per 5 seconds. (The reason for the throttling is that, although snapshotting is cheap, serializing a a snapshot to a string is expensive). If you only want to persist parts of the store you can use the `filter` option to filter which keys that should be stored.
 
 Options:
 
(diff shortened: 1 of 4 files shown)


zenflow:
@reekris What is the use case for this? Would it be more generally useful to accept a function, which can filter or transform the data in any way, rather than accepting an array of object paths?

reekris:
In my case I only want to store a users logged in token and some user settings and not the entire store. A problem can arise if the entire store is persisted and a key changes, then the entire cache is discarded when loaded. I my case that would mean that the users logged in state is lost.  

When looking at this I found a TODO comment about adding a key filter in the localStorageMixin file, so thats the way I implemented it. But I like your idea of supplying a function instead for more flexibility, maybe at the cost of ease of use. Do you want me to adjust the Pull request with this change?

zenflow:
> Do you want me to adjust the Pull request with this change?
@reekris I'm not actually using this package or looking after it anymore, so that's not really for me to say. I just thought I would contribute my two cents :)

I'm still not sure if anyone would ever need to filter or transform the data in any *different* way than this (`pick`ing only certain key paths), so maybe this way is better since it's easier to use 🤷

reekris:
@chrisdrackett May I ask why this issue was closed? This functionality is mentioned as a TODO in the source code: https://github.com/mobxjs/mst-gql/blob/104d0a9c237fbb6b34287442936c69928b1d6760/src/localStorageMixin.ts#L11

chrisdrackett:
I don’t remember closing this so it must have been a mistake

chrisdrackett:
I’m really confused how this happened. I wasn’t even awake when in theory I closed this :/

chrisdrackett:
oh, I see what happened. We renamed the `master` branch `main` and when the old master branch got deleted all the PR's got closed. Please re-open this, and sorry, that's my bad!

PR #246, opened 2020-06-08
Generate queries with primitives

### LINKED ISSUE #245: support for 'simple' return types
Right now if a query or mutation returns something other than an object we skip it. It would be nice to add these to the root store and just directly return the result as `data` from useQuery or similar.

chrisdrackett:
fixes #245

### DIFF
--- a/examples/1-getting-started/src/server/schema.js
+++ b/examples/1-getting-started/src/server/schema.js
@@ -23,6 +23,7 @@ const store = {
 const typeDefs = `
   type Query {
     todos: [Todo]
+    stringFromServer(string: String): String
   }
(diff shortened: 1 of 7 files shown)


examples/1-getting-started/src/server/schema.js
```diff
@@ -45,6 +46,9 @@ const resolvers = {
   Query: {
     todos: (root, args, context) => {
       return store.todos
+    },
+    stringFromServer: (root, { string }, context) => {
+      return string ? string : "No String Sent."
```
special-character, line 51:
```suggestion
      return string || "No String Sent."
```

PR #251, opened 2020-06-20
Add optional filter option to localStorageMixin 

chrisdrackett:
Copied from #243

Allows the user to filter which data to store in local storage

Uses lodash/pick since it allows to use dot notation for filtering nested data which is nice. But maybe there's a better alternative?

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,13 +633,14 @@ For examples, see the sections [Loading and rendering your first data](#loading-
 
 ## `localStorageMixin`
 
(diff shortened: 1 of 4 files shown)


reekris:
@chrisdrackett Is there anything I can do to move this forward and get it merged? Do you want it to be solved in another way?

chrisdrackett:
I haven't really looked at it, I know @zenflow had comments on the previous PR. I'm happy to take a look, but I'm not sure when I'll have time

Issue #255, opened 2020-07-21
Root Store Base references model that isn't generated for unions


special-character:
I am currently working on a search feature in my application that requires a union type. The ending schema.graphql that I am scaffolding from looks like:

```
type Item {
  id: ID
  title: String
  type: ItemType!
  ...
}

type Project {
  id: ID
  items: [Item!]!
  title: String!
  ...
}

union SearchResult = Item | Project
```

The `RootStore.base` references a `SearchResultModelType` in an action but a model isn't generated for union types. After taking a look at the generation code, I think this is by design: https://github.com/mobxjs/mst-gql/blob/104d0a9c237fbb6b34287442936c69928b1d6760/generator/generate.js#L312

I think we might need to figure how create the type that is expected in the action instead of having `SearchResultModelType`. It makes sense to me that this would be a union on the frontend as well in TS, something like `type SearchResult = Item | Project`

I can plan on taking a look at this use case more since it is my own. I still need to do a little digging to see what is generating the reference to `SearchResultModelType` in RootStoreBase and figure out how the unions are generated.

PR #257, opened 2020-07-21
Union type in actions

special-character:
This PR aims to solve https://github.com/mobxjs/mst-gql/issues/255.

- Adds code generation in the `<type>ModelSelector` file for unions (unions don't generate an actual Model or Model.base file and I was following the pattern of putting the extra stuff in this file)
- Adds imports to the `<type>Union` to the `RootStore.base.ts` file for use in actions
- Adds generation to the `queryHelper` function to use the correct type if the action is using a field that is `UNION`

In order to test this out, I added an example `search` field in the twitter example (only server side right now). It uses a union type of User and Message to return results, which creates a new action in the `RootStore.base.ts`. I also tested out the generation in another project (my own) which also generates correct types.

### DIFF
--- a/examples/3-twitter-clone/src/app/models/SearchResultModelSelector.ts
+++ b/examples/3-twitter-clone/src/app/models/SearchResultModelSelector.ts
@@ -3,8 +3,12 @@
 /* tslint:disable */
 
 import { QueryBuilder } from "mst-gql"
+import { MessageModelType } from "./MessageModel"
 import { MessageModelSelector, messageModelPrimitives } from "./MessageModel.base"
+import { UserModelType } from "./UserModel"
 import { UserModelSelector, userModelPrimitives } from "./UserModel.base"
+
+export type SearchResultUnion = UserModelType | MessageModelType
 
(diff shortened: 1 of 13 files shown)


examples/3-twitter-clone/src/app/models/SearchResultModelSelector.ts
```diff
@@ -0,0 +1,22 @@
...
+/* eslint-disable */
+/* tslint:disable */
+
+import { QueryBuilder } from "mst-gql"
+import { MessageModelType } from "./MessageModel"
+import { MessageModelSelector, messageModelPrimitives } from "./MessageModel.base"
+import { UserModelType } from "./UserModel"
+import { UserModelSelector, userModelPrimitives } from "./UserModel.base"
+
+export type SearchResultUnion = UserModelType | MessageModelType
```
special-character, line 11: Only additional generated code for unions is in this line

special-character:
I actually think something might be off with this. Using it in my own code and there is an issue with the fragment selector generation.

Issue #223, opened 2020-05-02
Support Multiple fields with type `ID`


chrisdrackett:
right now we only expect a single ID field. However, in the case a user has more than one ID field on an object mst-gql will not know which one is the ID for the object.

This issue covers updating the config or similar to support letting mst-gql know which is the canonical ID for any model that has more than one ID.

We should provide a clear error during scaffold that we ran into this and give the users steps to fix it.

mwarger:
Something additional to consider: not all schemas use `ID` as the type for ID (Hasura, for example, could have a `uuid` as the ID type).

chrisdrackett:
I'm confused, they use something other than `ID` in the graphQL schema? what is the logic behind this?

barbalex:
@mwarger oh, is that why I always get
```
Uncaught Error: Failed to resolve reference 43e8a390-9447-11ea-8ccb-19e3a3dddd88 to late(function () {
    return _gartenModel__WEBPACK_IMPORTED_MODULE_11__["gartenModel"];
  })
```
when trying to select fields in referenced tables?

My project uses hasura.

barbalex:
related issue: https://github.com/mobxjs/mst-gql/issues/215

barbalex:
Actually there is an issue just about hasura not using type `ID`: https://github.com/mobxjs/mst-gql/issues/92

chrisdrackett:
good point, I feel like we could fix both these issues with a single fix of optimally specifying an ID via config.

/cc @beepsoft

Metevier:
@chrisdrackett 
I am interested in implementing this if it's a feature we feel like is still needed.

How do you feel about the proposal in this comment?
https://github.com/mobxjs/mst-gql/issues/92#issue-493537797

chrisdrackett:
@Metevier I'm all for this! We don't run into this ourselves, so I don't have a lot of ideas around the best way to handle it, but seems like a place where we can add some optional flexibility!

Issue #92, opened 2019-09-13
Alternative way to generate types.identifier/types.identifierNumber for non ID typed identifier fields


beepsoft:
The problem
----
Currently - and very logically - the generator generates `types.identifier` MST type for graphql fields of type `ID`. For example:

```
type Todo {
  id: ID
  ...
}
```

becomes:

```
export const TodoModelBase = MSTGQLObject
  .named('Todo')
  .props({
    __typename: types.optional(types.literal("Todo"), "Todo"),
    id: types.identifier,
    ...
  })
```

Hasura (https://github.com/hasura/graphql-engine), however, generates schemas for PostgreSQL tables matching the type of the columns in the database. So, in case my id in the table is of type `bigint`, which is quite typical for a primary key, the generated schema will look like this:

```
scalar bigint

type Todo {
  id: bigint!
  ...
}
```

And for this mst-gql generates:

```
export const TodoModelBase = MSTGQLObject
  .named('Todo')
  .props({
    __typename: types.optional(types.literal("Todo"), "Todo"),
    id: types.maybeNull(types.frozen()),
    ...
  })
```

As you can see the type of `id` becomes `types.maybeNull(types.frozen())`. The problem with this is that resolving references in the state tree won't work for identifiers that are not defined as `types.identifier` (or `types.identifierNumber`).

What I would need is something like:

```
export const TodoModelBase = MSTGQLObject
  .named('todo')
  .props({
    __typename: types.optional(types.literal("Todo"), "Todo"),
    id: types.identifierNumber,
    ...
  })
```

_Note that I need `types.identifierNumber` and not even `types.identifier` which is what mst-gql is able to generate currently._

Proposal
----

So, my proposal is to introduce a mechanism, which can recognize identifiers by name/pattern matching in the graphql schema beside those that have the graphql type `ID`.

Something like:

`mst-gql --overrideFieldType "id:bigint:types.identifierNumber,*_id:bigint:types.identifierNumber"`

The `--overrideFieldType` would get a specification of field names to match and the kind of mst type definition to generate for that type instead of the default type. In the above case fields that are called `id` or fields that end with `_id` and are of graphql type `bigint` would be defined as `types.identifierNumber` MST types.

The override specification may also have the field as Type.field, in which case only the field of that specific type would be overriden:

`mst-gql --overrideFieldType "Todo.id:bigint:types.identifierNumber`"

The graphqlql type may be omitted or also pattern matched as well:

[...]

PR #259, opened 2020-07-27
Add fieldOverrides to enable the usage of multiple and custom ID types

Metevier:
@chrisdrackett 
Attempts to add workarounds for #223 and #92 

Adds two new scaffolding options:
`--useIdentifierNumber`: Specifying this option will use `identifierNumber` instead of `identifier` for all GQL ID types project wide. 
`--fieldOverrides gqlFieldName:gqlFieldType:mstType,...`: draws largely from the proposal in [this comment](https://github.com/mobxjs/mst-gql/issues/92#issue-493537797) The explanation in the README is hopefully sufficient, but it may need to be tweaked as its pretty verbose. The explanation is as follows:

`--fieldOverrides id:uuid:idenfitier,*:ID:identifierNumber` Overrides default MST types for matching GraphQL names and types. The format is `gqlFieldName:gqlFieldType:mstType`. Supports full or partial wildcards for fieldNames, and full wildcards for fieldTypes. Case Sensitive. If multiple matches occur, the match with the least amount of wildcards will be used, followd by the order specified in the arg list if there are still multiple matches. Some examples:

  - `*_id:*:string` - Matches any GQL type with the field name `*_id` (like `user_id`), and uses the MST type `types.string`
  - `*:ID:identifierNumber` - Matches any GQL type with any field name and the `ID` type, and uses the MST type `types.identifierNumber`
  - `User.user_id:ID:number` - Matches the `user_id` field on `User` with the GQL type `ID`, and uses the MST type `types.number`

    Specifying this argument additionaly allows the use of multiple IDs on a type. The best matched ID will be used, setting the other IDs to `types.frozen()`

  - `Book.author_id:ID:identifierNumber` - Matches the `author_id` field on `Book` with the GQL type `ID` and uses the MST type `types.identifierNumber`, and sets any other GQL IDs on `Book` to `types.frozen()`

    For TS users, input types and query arguments will only be modified for fieldOverrides with a wildcard for `gqlFieldName` (`*:uuid:identifier`). An override like `*_id:uuid:identifier` will not affect input types.

    The primary use case for this feature is for GQL Servers that don't always do what you want. For example, Hasura does not generate GQL ID types for UUID fields, which causes issues when trying to reference associate types in MST. To overcome this, simply specify `--fieldOverrides *:UUID:identifier`

My main concerns around this PR:
1. The logic to determine which match has the highest precedence isn't the easiest to follow.
2. How to handle TS for this problem isn't well defined, so the current implementation only overrides TS types for overrides that specify a wildcard for GQLType, like `*:uuid:identifier`. Usage of this is feature is needed to determine what the final functionality should be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,7 @@ The `mst-gql` command currently accepts the following arguments:
 - `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
 - `--dontRenameModels` By default generates model names from graphql schema types that are idiomatic Javascript/Typescript names, ie. type names will be PascalCased and root collection names camelCased. With `--dontRenameModels` the original names - as provided by the graphql schema - will be used for generating models.
+- `--useIdentifierNumber` Specifies the use of `identifierNumber` instead of `identifier` as the mst type for the generated models unique keys. This requires your models to use numbers as their identifiers. See the [mobx-state-tree](https://mobx-state-tree.js.org/overview/types#property-types) for more information.
 - `source` The last argument is the location at which to find the graphQL definitions. This can be
   - a graphql endpoint, like `http://host/graphql`
   - a graphql files, like `schema.graphql`
--- a/README.md
+++ b/README.md
@@ -409,7 +409,21 @@ The `mst-gql` command currently accepts the following arguments:
 - `--noReact` doesn't generate the React related utilities
 - `--force` When set, exiting files will always be overridden. This will drop all customizations of model classes!
 - `--dontRenameModels` By default generates model names from graphql schema types that are idiomatic Javascript/Typescript names, ie. type names will be PascalCased and root collection names camelCased. With `--dontRenameModels` the original names - as provided by the graphql schema - will be used for generating models.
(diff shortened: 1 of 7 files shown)


chrisdrackett:
sorry for the delay in reviewing this PR. I haven't had a lot of time recently and this functionality is new to me. I'll throw out some review requests to see if we can get any movement here.

jesse-savary:
I'm not confident that endless configuration flags are the solution to issues with mapping GraphQL to MST;
I feel that a true `mst-gql@1.0.0` needs a more robust solution. That being said, this provides workarounds for several issues that people (myself included) have been running into. Will review and merge in the next couple days.

jesse-savary:
Attempting to merge this but admittedly my Git skills aren't the best, having some issues getting things to merge cleanly. @chrisdrackett any resources/general tips? I'm a bit unfamiliar with the workflow of bringing an older PR up to date via a rebase + getting it ready to merge; don't want to pollute the repository.

chrisdrackett:
I wouldn't worry about rebasing. I would just checkout this PR branch and then pull in `main` and resolve issues locally. You could also resolve directy on GitHub using the `Resolve conflicts` button on the PR

PR #264, opened 2020-08-08
Union ofTypes issue fixed

### LINKED ISSUE #263: Interface and Union .ofTypes error
Hello Dev,

I'm using this product in couple of my developments. Great tool 👍.

Well I ran into following error recently while scaffolding a Union Type.

```
.\node_modules\mst-gql\generator\generate.js:503
      const mstUnionArgs = interfaceOrUnionType.ofTypes.map(t => {
                                                ^

TypeError: Cannot read property 'ofTypes' of undefined
    at handleInterfaceOrUnionFieldType (.\node_modules\mst-gql\generator\generate.js:503:49)
    at handleFieldType (.\node_modules\mst-gql\generator\generate.js:458:25)
    at handleField (.\node_modules\mst-gql\generator\generate.js:407:34)
Compiled with warnings.
```

I tried to look into the issue, then I found a bug with name conventions for union types. `type.names` are converted but while getting the union types it is comparing `possibleTypes.name` which is not processed with naming convention. Which results into following issue.

So I have fixed it and added a pull request, review the changes and add it into further releases.

Thanks

roshanreacts:
#263 
Fixes #263

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -77,10 +77,10 @@ export const ModelBase = MSTGQLObject
 
   function generateTypes() {
     types
-      .filter((type) => !excludes.includes(type.name))
-      .filter((type) => !type.name.startsWith("__"))
-      .filter((type) => type.kind !== "SCALAR")
-      .forEach((type) => {
+      .filter(type => !excludes.includes(type.name))
+      .filter(type => !type.name.startsWith("__"))
+      .filter(type => type.kind !== "SCALAR")
+      .forEach(type => {
         knownTypes.push(type.name)
         if (type.kind === "OBJECT") {
           origObjectTypes.push(type.origName)
@@ -96,7 +96,7 @@ export const ModelBase = MSTGQLObject
       )
     }
 
-    rootTypes.forEach((type) => {
+    rootTypes.forEach(type => {
       if (!origObjectTypes.includes(type)) {
         if (isTypeReservedName(type)) {
           throw new Error(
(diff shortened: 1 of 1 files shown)


special-character:
@roshanreacts Thanks for the PR, would you be willing to add to one of the examples so we could see this running? I don't think we have an example that runs into the problem this solves and it would be nice to be able to test it out with the same kind of schema that you have from #263 

Example 3 might be a good option since it has a Union type already.

chrisdrackett:
@roshanreacts converting this to a draft until we can get an example or test up and running.

roshanreacts:
Hi @chrisdrackett  and @special-character ,

I'm not sure with name conventions of GraphQL. But this issue is triggered when you have name of the schema/type with capital letters. Something like `type AWS` or `type PRComment`.

Is there any suggestion in GraphQL that we cannot have such name for types?

special-character:
@roshanreacts I think having types with capital letters is ok, but I can't seem to repro the issue. All of the types we use start with capitals and we have an example with a union here:

https://github.com/mobxjs/mst-gql/blob/main/examples/3-twitter-clone/src/server/schema.graphql#L29

If I run scaffold for example 3, we don't run into the same issue. I was hoping to repro and write a test for the failing case so we don't goof it again if we rewrite the generation code in the future (🤞we can at some point to make it a tad easier to change)

roshanreacts:
Hi @special-character ,

Sorry that I was not clear not just starting with capital letter. Try with at least 2-3 Capital letters in the beginning.

`type ABCdef`

special-character:
@roshanreacts aaah, thanks! I didn't catch that distinction. I will try that tomorrow morning.

special-character:
Sorry I didn't get to this today, I will be out for a week but will follow up on repro/tests when I get back

special-character:
Just tried this out and it seems to scaffold fine if I do something like:

```
type BBob {
  name: String
}
union SSaget = User | Message
```

I am still not entirely sure if this is still an issue?

roshanreacts:
Seriously... Can I have a look into the generated files somewhere like codesandbox.

special-character:
I am running this on example 3 which uses the latest version of mst-gql.

jesse-savary:
Will attempt to reproduce this.

Issue #273, opened 2020-09-02
Add a clear or reset function to query


mattiasewers:
I'm in need of a clear or reset function after a query has loaded data, I'll make a pr but need feedback of what it should be called. The functionality would just be able to set `data` to `undefined`

PR #275, opened 2020-09-03
Add clear method to Query

mattiasewers:
This is related to the issue #273 which adds a clear method to be able to reset the query to undefined again after fetching.

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -93,7 +93,7 @@ export class Query<T = unknown> implements PromiseLike<T> {
     }
   }
 
-  clear = (): void => {
+  clearData = (): void => {
     action(() => {
       this.data = undefined
     })
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -93,6 +93,12 @@ export class Query<T = unknown> implements PromiseLike<T> {
     }
   }
 
+  clear = (): void => {
+    action(() => {
+      this.data = undefined
+    })
+  }
+
   refetch = (): Promise<T> => {
     return Promise.resolve().then(
       action(() => {


src/Query.ts
```diff
@@ -93,6 +93,12 @@ export class Query<T = unknown> implements PromiseLike<T> {
     }
   }
 
+  clear = (): void => {
```
chrisdrackett, line 96: my only potential suggestion here is using `clearData` vs. clear just to make it more clear what you are clearing. clear?

chrisdrackett:
@mattiasewers just one comment, but also curious what use case you are hitting with this

special-character:
@mattiasewers Thanks for the PR. Could you explain a bit about what this is helping you solve? I understand you are clearing the data for the Query but I don't quite understand why you need to do this.

mattiasewers:
@chrisdrackett @special-character  I'm working on a app for my company where I need to reset the data object without making a another request, I don't know if that is possible currently which is why I made this pr.

src/Query.ts
mattiasewers: Ok I thought that was clear but we can rename it.

special-character:
@mattiasewers Thanks for commenting, I get that you need to clear the data I was curious about what the use case was and how you were using the `Query`. I haven't used the data directly from the `Query` object because it has been more internal for `mst-gql` (until now). The only case I was thinking of was for clearing the cache for the query the next time it is called with the same args. I always use the data out of the store or use the query options to change how a particular request acts.

Do you need it to clear the cache or is there another use case that you need this for?

mattiasewers:
@special-character Yes I need to clear the cache

src/Query.ts
chrisdrackett, line 96:
```suggestion
  clearData = (): void => {
```

PR #288, opened 2020-10-06
Fix interface generation/references 

### LINKED ISSUE #256: Interfaces from graphql schema not generated into models, but used in rootStore actions
I have a schema with interface:

```
type Query {
    available(input: AvailableInput): [Element!]
}
input AvailableInput {
    elementID: ID!
}

interface Element {
    payload: Payload!
}

type FixedElement implements Element {
    payload: Payload!
    value: String!
}

type NumberElement implements Element {
    payload: Payload!
    amount: String!
}

type Payload {
    value: ID!
}
```

After generating I have an error in rootStore.base.ts with using `ElementModelType` but its not exists.

Repo with error reproduction: [https://github.com/ivkirill/mst-gql-interface ](url)

special-character:
closes #256 

This effectively does the same thing we have done for unions in the past: https://github.com/mobxjs/mst-gql/pull/257

- Adds type export from the interface selector file (full model files are not generated for abstract types)
- RootStore.base now imports correct selectors and primitives from the selector file and uses them in the actions

In this PR I also updated tests to expect the new generated code as well as fixing the tests running by putting the `.babelrc` file back.


.babelrc
```diff
@@ -0,0 +1,10 @@
+{
```
special-character, line 1: I added this file back in so jest tests would succeed. I had to take out one extra line that rollup was complaining about: https://github.com/mobxjs/mst-gql/pull/288/commits/72495b3aedd08233aa55e09d2c2e2ce1c9588edd

I tried to use [rollup-jest](https://github.com/ambar/rollup-jest) but tests hang and never succeed.

jesse-savary:
Looks good! I'll sort out the conflict and get this merged, thank you.

PR #298, opened 2020-12-21
Fix for circular types on relation fields

### LINKED ISSUE #297: Relation types in union cause circular references
Hello!

Typescript support for my mst-gql project broke due to circular type references after adding a union type on a relationship field in my schema.

This is the line in `MyModel.Base.ts` that appears to cause the problems.

```
    // Broken:
    item: types.union(types.undefined, types.null, types.union(types.late(() => SomeItemType), types.late(() => AnotherItemType))),
```

Per #267, relation fields are normally be specified like `types.late((): any => SomeItemType)`

I confirmed that adding the `: any` fixes the circular type references in my code:

```
    // Fixed:
    item: types.union(types.undefined, types.null, types.union(types.late((): any => SomeItemType), types.late((): any => AnotherItemType))),
```

emckay:
Fixes #297.

This resolves the issue in my codebase, but I'm very unconfident in this PR because I admittedly do not understand why there is the `isSelf` check before applying the `: any`. Feel free to close this if there's a better fix.

Thank you!

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -510,10 +510,9 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
           // TODO: split field type resolvement from model properties output
           addImport(subTypeClassName, subTypeClassName)
         }
-        const isSelf = fieldType.name === currentType
         // always using late prevents potential circular dependency issues between files
         return `types.late(()${
-          isSelf && format === "ts" ? ": any" : ""
+          format === "ts" ? ": any" : ""
         } => ${subTypeClassName})`
       })
       return `types.union(${mstUnionArgs.join(", ")})`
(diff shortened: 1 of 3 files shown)


emckay:
Actually, this causes the union property to get type `unknown`, so this is probably not the right approach or additional work is required to make this behave the same way as relation fields that a union.

emckay:
Would someone be able to provide some guidance on this? We're using this patch but it is causing problems because the union types are getting typed as `unknown` when they're being used later. Happy to create a minimal reproduction of the issue if that would be useful.

jesse-savary:
Will review this soon.

jesse-savary:
@emckay could I get the minimal reproduction of this?

emckay:
Sorry for slow response. Will try to put something together by the end of next week. Thanks!

emckay:
I was not able to reproduce this easily. We have a lot of types with references to each other and long cycles of circular dependencies (e.g. Todo -> TodoList -> User -> Projects -> TodoList -> Todo), and it seems like it will take to long to figure out the exact chain of dependencies that causes this issue.

jesse-savary:
Hmmm, my project also has issues with circular types. Wondering what the ultimate solution is?

Issue #302, opened 2021-01-14
Looking for maintainers!


chrisdrackett:
Within the last 6 months we've been moving away from using `mst-gql` in our project and as a result I've been spending less and less time maintaining this project. While I'm not the sole maintainer, I have been doing a good amount of the work around preparing and doing releases, which clearly has fallen off its normal pace.

Hopefully either an existing maintainer or someone new is able to step up and manage releases. I'm happy to support in the transition as much as possible!

mattiasewers:
@chrisdrackett Hi, why did you move away from it and what are you using instead?

rdewolff:
Sorry to read this. That's the kind of issue I don't like to read when considering to use mst-gql for our saas app.

Could you please elaborate a bit more on:

> We want our app to have great offline support and there are just too many little issues with our current approach.

That'd help the community :)

beepsoft:
@chrisdrackett I'm not really familiar with the npm publishing workflow yet, but I worked a bit on mst-gql before and I would like to see the project develop and succeed. I use it in a number of react SPA and react-native apps, so I would be glad to help with the maintenance if someone could join me in this.

beepsoft:
@chrisdrackett did you succeed finding a maintainer? I've seen @jesse-savary merged a PR recently. Will he manage further developments? I'm really looking forward to #294 and #303 be merged.

jesse-savary:
@beepsoft will deal with these today; `mst-gql` is very important to my company/project and I'm behind it full-force.

I apologize for not paying better attention to the open PRs. Please feel free to tag me or assign me where needed. Let's get this train moving 😄

beepsoft:
@jesse-savary it is really good news, thanks for your work! 😃 I'm also very much invested into `mst-gql` so I hope it will get better and better.

chrisdrackett:
Thanks so much @jesse-savary! let me know if you need anything from me or have any questions!

jesse-savary:
My current goal is to get every (valid) PR merged in and to take care of any small issues. @chrisdrackett I will likely ping you to sort out releases + ensure I don't blow up NPM; expect to hear from me sometime in the next week or two :)

chrisdrackett:
sounds good!

mylifenp:
I have been trying mst-gql from few months now. I find it really fascinating. I am still probably a noob,  but would really love to help the project grow. Is there anything i can do?

beepsoft:
[...]

PR #303, opened 2021-01-18
Expose Clients for later initiation

BradenM:
This PR exposes `gqlHttpClient` and `gqlWsClient` by moving them to `MSTGQLStore`'s volatile state and providing setter actions.

This is so either client can be initiated / changed at a later time instead of being de-referenced from the mst env.

Personally needed this because my env had an initial state that defined `gqlWsClient` as null and later initiated it when it could be provided with an authentication token. But, since the current behavior de-references the clients, reassigning it from my env would not propagate to mst-gql and it would throw an error.

### DIFF
--- a/src/MSTGQLStore.ts
+++ b/src/MSTGQLStore.ts
@@ -23,32 +23,28 @@ export const MSTGQLStore = types
     ssr: boolean
     __promises: Map<string, Promise<unknown>>
     __afterInit: boolean
+    gqlHttpClient: RequestHandler
+    gqlWsClient: SubscriptionClient
   } => {
     const {
-      ssr = false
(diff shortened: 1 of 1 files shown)


beepsoft:
This is really great, I've also had problems reconfiguring clients - especially the WS client - when authentication changes.  @chrisdrackett can you please take a look at this?

chrisdrackett:
@beepsoft I'm probably not going to be able to get to looking at this as we're no longer using this library and my free time is extremely limited (https://github.com/mobxjs/mst-gql/issues/302)

jesse-savary:
Looks good, thanks a lot @BradenM

PR #315, opened 2021-03-12
Fix broken snaphsot

### LINKED ISSUE #313: Rehydrating state from snapshot breaks type inference on union types
I am starting to use SSR in my application and ran into this issue when rehydrating the state from a snapshot.

If your schema looks like:

```
  union Todo = BasicTodo | FancyTodo
  type BasicTodo {
    id: ID,
    text: String,
    complete: Boolean,
  }
  type FancyTodo {
    id: ID,
    label: String,
    color: String,
    complete: Boolean
  }
  type TodoList {
    id: ID!
    todos: [Todo!]!
  }
```

when TodoList.todos is populated and attempted to restore from a snapshot, MST will only look in the store for the first type of the union type to find the object, resulting in a run-time crash.

Reproduction: https://github.com/mobxjs/mst-gql/commit/0a320033d398f74c1ccee722901cca6fc8216621

Happy to help on this with some guidance.

emckay:
Fixes #313

This PR adds a test suite for union types and includes a fix for #313 (bug when applying snapshot with array of union types).

THIS PR MAY BE CONSIDERED A BREAKING CHANGE. SO PLEASE BE CAUTIOUS IF YOU MERGE THIS.

I change the ids of the referenced objects to include a label MSTGQLRef. This label allows library users to correctly discriminate union types, like this:

```js
      types.union(
        {
          dispatcher: (ss) => {
            const modelName = ss.__typename
              ? ss.__typename
              : getMSTGQLRefLabelAndId(ss).label            
            if (modelName === "FancyTodo") {
              return MSTGQLRef(
                types.late(() => FancyTodoModel),
                "FancyTodo"
              )
            } else {
              return MSTGQLRef(
                types.late(() => BasicTodoModel),
                "BasicTodo"
              )
            }
          }
        },
        //...
     )
 ```

However, changing this id may break existing deployments that use persisted snapshots with MSTGQLRefs, especially if the delimiter I chose (`::`) is used in the ids.


gsdean:
Adding a magic string that isn’t configureable is a pretty big gotcha. Especially since :: is a common deliminator in many languages. 

Would you entertain a PR making this configureable?

PR #319, opened 2021-04-02
Fix 'id' getter MSTGQLRef 

weglov:
This PR contains a small fix for getting the node.id in MSTGQLRef.

**Problem:**
When I create my mobxStore from a snapshot during testing, the models do not call a setter where their id must be modified and a separator added.
```
RootStore.create({
   items: {
       '1': {id: '1', name: 'one'},
       '2': {id: '2', name: 'two'}
   },
   myRef: ['1'] //  MSTGQLRef(types.late((): any => ItemModel)),
})
```
> Also, you can look real example in test https://github.com/mobxjs/mst-gql/pull/319/files#diff-8da68bbe119c10527a005b652d591c418c435a78af66333be4e4608e5fc3cc72 

I got:  `Failed to resolve reference   to late(() => ItemModel.ItemModel)`, as you can see without id becouse id in this case just empty space `''`

So, this happens because in this case not fires this code: https://github.com/mobxjs/mst-gql/blob/main/src/MSTGQLObject.ts#L41-L43

And my solution will fix that, although I'm not sure if it's the right solution, but it works for me and theoretically can't cause problems.

### DIFF
--- a/src/MSTGQLObject.ts
+++ b/src/MSTGQLObject.ts
@@ -15,8 +15,12 @@ import { StoreType } from "./MSTGQLStore"
 */
 const MSTGQL_ID_DELIM = "::"
 export function getMSTGQLRefLabelAndId(labeledId: string) {
-  const [label, ...id] = labeledId.split(MSTGQL_ID_DELIM)
-  return { label, id: id.join("") || labeledId }
+  if (labeledId.includes(MSTGQL_ID_DELIM)) {
+    const [label, ...id] = labeledId.split(MSTGQL_ID_DELIM)
+    return { label, id: id.join("") }
+  }
+
+  return { label: null, id: labeledId }
 }
 
 export function MSTGQLRef<T extends IAnyModelType>(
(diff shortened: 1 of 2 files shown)


weglov:
@emckay hi man, I would be happy if you would watch this PR because they are closely related to your changes.

emckay:
Thanks @weglov.. Will take a look this afternoon!

emckay:
Thanks for debugging this and sorry about introducing the breaking change!

Right now, the problem you're running into is that if there's no delimiter in the id, then `getMSTGQLRefLabelAndId` returns `{label: 'something', id: ''}`. I think that ideally we'd return `{ label: null, id: 'something' }`. This PR returns `{ label: 'something', id: 'something' }`

Would you mind updating your PR to return `label: null` if there's no delimiter in the string? Happy to take a stab at it if you don't have time!

emckay:
Looks good! @jesse-savary do you mind reviewing + merging this?

jesse-savary:
Looks good, thank you @weglov

weglov:
@jesse-savary I will be happy if you published new version:) 14.0.1

jesse-savary:
Oops, still getting the hang of this. Will cut a release shortly!

jesse-savary:
Published @weglov :)

weglov:
Thank you so much! Really appreciate it!

jesse-savary:
@weglov resolved, apologies for the delay! I accidentally published `0.14.1` to my private registry instead of NPM :)

PR #320, opened 2021-04-06
Add safe typing for mst late

ivkirill:


jesse-savary:
Hey @ivkirill, just wondering what's up with this PR? Safe typing of mst-gql is a goal of mine but I wasn't sure if any of this code was worth testing/iterating on.

ivkirill:
@jesse-savary tested for pet project, but inner test with circular deps still unresolved

jesse-savary:
@ivkirill Understandable, hopefully I will get time in the coming weeks to find a better solution.

PR #348, opened 2021-11-23
Fix handling of union of root types

### LINKED ISSUE #304: GraphQL union type fields aren't created as references
If a model contains a Union type object, we end up with
`types.union(types.late(() => AreaModel), types.late(() => ScheduleModel))`

when it should actually be
`MSTGQLRef(types.union(types.late(() => AreaModel), types.late(() => ScheduleModel)))`

Because it's not creating it as a reference, it will fail to load the snapshot, as it will try to create a duplicate object of the Union'd type instead of referencing the existing object in the RootStore map.

I'm guessing this is may be because only roots are turned into MSTGQLRef's elsewhere, and union types of course not being turned into roots

smokku:
Properly generate `MSTGQLRef()`s for union type fields.

Fixes #304

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -467,8 +467,10 @@ ${generateFragments(name, primitiveFields, nonPrimitiveFields)}
       }
     }
 
-    function handleObjectFieldType(fieldName, fieldType, isNested) {
-      nonPrimitiveFields.push([fieldName, fieldType.name])
+    function handleObjectFieldType(fieldName, fieldType, isNested, skipField) {
+      if (!skipField) {
+        nonPrimitiveFields.push([fieldName, fieldType.name])
+      }
       const isSelf = fieldType.name === currentType
 
       // this type is not going to be handled by mst-gql, store as frozen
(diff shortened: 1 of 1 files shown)


jesse-savary:
Hey @smokku, thank you for this. I had to pass `null` for `fieldArgs` when calling `handleObjectFieldType` as I merged another of your PRs. I'm planning on doing a bunch of testing this week with mst-gql before cutting a new release but I was wondering if you could help out specifically with this fix so I can verify I didn't break it :)

smokku:
Just checked my local `generate.js` and this is the proper change, with both PR merged, with one difference:
```js
return handleObjectFieldType(t.name, t, fieldArgs, false, true)
```
but `null` works too. ;-)

smokku:
For completeness, here is the full diff og `generate.js` I use on production: https://gist.github.com/smokku/5df4400e898d9805f45ddbb3af1180c5

jesse-savary:
Great -- I'll get this patched up before release, thank you!

adumat:
@smokku thanks a lot!

PR #350, opened 2021-11-23
Call apollo binary by relative path

smokku:
Do not assume packages are nested.
Package managers such as `pnpm` are flattening the package directory.


iswara108:
This doesn't work with Yarn 2 PnP because PnP doesn't have "node_modules/".

jesse-savary:
#353 also mentioned issues with PnP; I'm going to see if I can solve the issue for both configurations.

jesse-savary:
Closing this because we no longer use the Apollo CLI.

PR #370, opened 2022-05-22
feat: add custom scalars support

dmytro-shpak:


dmytro-shpak:
It looks like we should not create the new property

PR #375, opened 2022-06-13
Add support of type import to --fieldOverrides

dmytro-shpak:
One more fragment in fieldOverrides allows to import type definition
from file. May be used for custom scalars support with MST type.custom()


jesse-savary:
Hey @dmytro-shpak, I've added your commits into https://github.com/mobxjs/mst-gql/pull/382 because the `development` branch is more up to date than `main`. Please give version `0.17.0` a go and let me know what you think.

dmytro-shpak:
Hi @jesse-savary, thank you for your reaction.  I really like this project and could spend some time on it. It work on mst-gql coordinated somehow?

PR #378, opened 2022-06-28
Scaffold nullable types as "(type | null)"

smokku:
Nullable types are generated as optional in typescript files. This makes the fields of `type | undefined`.

 When you try to set `null` as such field, TS compiler refuses to assign `null` to optional field.

The main problem is that it prevents sending `null` to backend to erase field value in mutation.

Workaround is to do something like this
```ts
store.mutateSetSomething({
  value: null as unknown as undefined,
})
```

This PR fixes this by generating nullable types as `(type | null)`.

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -1032,7 +1032,7 @@ ${optPrefix("\n    // ", sanitizeComment(description))}
 
     return `${name}${
       canBeUndefined || fromUndefineableList ? "?" : ""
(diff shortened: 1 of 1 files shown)


jesse-savary:
Hey @smokku, I've added your commits into https://github.com/mobxjs/mst-gql/pull/382 because the `development` branch is more up to date than `main`. Please give version `0.17.0` a go and let me know what you think.

PR #381, opened 2022-07-14
V0.16.0 graphql ws subscribe

ChrisABryant:
This change adds graphql-ws support only for subscriptions. 

Is backwards compatible so still works with subscription-transport-ws without code changes.

Only difference is in new transport parameter for example:
```
      const gqlHttpClient = createHttpClient(httpUrl, {
         headers: {
            authorization: 'Bearer ' + token,
         }
      });

      const graphqlWsClient = createClient({
         url: wsUrl,
         connectionParams: {
            jwt: token,
         }
      });

      {
         gqlHttpClient,
         graphqlWsClient // instead of gqlWsClient
      }
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "editor.tabSize": 3,
   "eslint.enable": false,
   "cSpell.enabled": false,
   "typescript.tsdk": "node_modules/typescript/lib"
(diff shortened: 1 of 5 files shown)


jesse-savary:
Hey @ChrisABryant, I've added your commits into https://github.com/mobxjs/mst-gql/pull/382 because the `development` is more up to date than `main`. Please give version `0.17.0` a go and let me know what you think.

PR #387, opened 2022-07-23
Support [String] as query return type

smokku:
Queries having return type of `[String]` get scaffolded as

```typescript
      queryFetchStuff(
        variables?: {},
        resultSelector: string | ((qb: StringModelSelector) => StringModelSelector) = stringModelPrimitives.toString(),
        options: QueryOptions = {}
      ) {
        return self.query<{ fetchStuff: StringModelType[] }>(
          `query fetchStuff { fetchStuff {
        ${typeof resultSelector === 'function' ? resultSelector(new StringModelSelector()).toString() : resultSelector}
      } }`,
          variables,
          options
        )
      },
```

treating `String` as a complex type. Obviously there is no `StringModelSelector` generated and build fails.

This change fixes that, generating proper query function.

```typescript
      queryFetchStuff(variables?: {}, options: QueryOptions = {}) {
        return self.query<{ fetchStuff: string[] }>(`query fetchStuff { fetchStuff }`, variables, options)
      },
```

### DIFF
--- a/generator/generate.js
+++ b/generator/generate.js
@@ -905,14 +905,14 @@ ${enumContent}
 
         let { name, origName, args, type, description } = field
 
-        const isScalar =
-          type.kind === "SCALAR" ||
-          (type.ofType && type.ofType.kind === "SCALAR")
-
         if (type.kind === "NON_NULL") type = type.ofType
         const returnsList = type.kind === "LIST"
         let returnType = returnsList ? type.ofType : type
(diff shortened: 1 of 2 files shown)


jesse-savary:
Could you update the test snapshots? (run `jest -u` in the root of the project)

smokku:
> @smokku could you update the test snapshots? (run `jest -u` in the root of the project)

Done.

PR #396, opened 2022-10-13
Fix MSTGQLStore.subscribe() when a DocumentNode is passed as query

beepsoft:
The MSTGQLStore.subscribe() fails when a DocumentNode is passed as query, ie. when an object created using graphql-tag is passed.

The problem is that query.toString() doesn't work here, because graphql doesn't implement it
to convert the DocumentNode to a graphql string representation. Instead, print(query) has to be used.

mst-gql has no tests for subscriptions, but this fix can be tested in a project using mst-gql using

```
    "mst-gql": "https://github.com/beepsoft/mst-gql#mst-gql-v0.17.1-subscription-fix-gitpkg",
```

in package.json


beepsoft:
Until this is merged the workaround is to use print()  in your code when calling subscribe. eg:

```js
import { print } from 'graphql/language/printer'

gqlStore.subscribe(print(gql`...some graphql query`), ...)
```

PR #398, opened 2022-10-25
feat: support of non standard id field name

dmytro-shpak:


dmytro-shpak:
The previous pull request was removed by accident. It looks like this feature need some unit tests before the merge

beepsoft:
Hi @dmytro-shpak! I try to figure what problem you try to solve here. Can you provide an example for it, or even better a test case?

PR #449, opened 2024-09-02
fix parse config fieldOverrides

alandell0n:
When the "fieldOverrides" configuration was passed in a config file (like "mst-gql.config.js") as an array, it was not working because it was not formatted with "parseFieldOverrides", needing to create a different value field like "[["*_id", "\*", "string"]]", when it should be just "["*_id:*:string"]".


beepsoft:
Thanks, @AlanDellonSchwartzman! Merged and fixed to check for undefined config.fieldOverrides. Published 0.17.3.

alandell0n:
Good observation. I'm glad I could contribute :)

